Issue #325, opened 2016-02-29
XENON100 S2(x,y) corrections


JelleAalbers:
This map should be converted from the Xerawdp format, I'll do this soon.

JelleAalbers:
Was added in #334.

PR #334, opened 2016-03-17
XENON100 S2(x,y) light yield + XENON100 simulation bugfix

JelleAalbers:
This includes the XENON100 S2(x,y) correction from Xerawdp (see #325). Here is a note to show how the conversion is done, and verify on simulated data that the result works:

https://github.com/XENON1T/XeAnalysisScripts/blob/master/mapmaking/Parse%20S2%20xy%20correction%20map.ipynb

For XENON100, I now enabled the `magically_avoid_dead_pmts` setting for the simulator. Since we have a relative light yield map, the simulator uses it to adjust the number of photons produced, which means  the number of photons is already reduced under a dead PMT. If we would then distribute photons normally, the photon count would be further reduced under dead PMTs.

This 'double counting' was probably the cause of some strange effects @sanderbreur saw while looking at the S1 corrections recently. We should be careful about this for XENON1T once we get LCE maps from data.

The zombie PMTs for S2s (see #237) could now be removed from the configuration. This was basically a hack to compensate for the lack of S2 LCE map. 

Some minor bugfixes:
  * Erroneous documentation in InterpolatingMap.py (leftover from previous implementation)
  * Mistaken calculation of correction map dimensionality. Didn't affect anything so far since other maps were actually edge cases (S1 was 3d = max dim, 0d was = min dim) for which the wrong code also works :-) (for S1 it's a bit more complicated to see why it would still work, it's due to z being the last position attribute you could get)


tunnell:
Went through it, looks fine.  Objections to merging before Monday?

sanderbreur:
Perfect, lets merge.

PR #337, opened 2016-03-29
Small fixes for XENON1T processing

JelleAalbers:
This includes a few small fixes for processing XENON1T data:

 * "Fix" (in a very dirty way) a race condition in the ROOT output during multiprocessing, where several threads tried to write and/or compile the event class cpp file at the same time;
 * Turn off 3d posrec for now, it's nonsense in gas (LCE map for liquid) anyway, and it slows the processing  by a factor 5-10.
 * Don't crash if pulses are completely outside the event . This indicates there is still (probably) an off-by-one error in the event builder... I though this protection was enabled before, but it probably got lost.


tunnell:
Fine, this is important for processing.

Issue #300, opened 2016-01-21
Consistent configuration file variables


pelssers:
There are currently some differences between XENON100 and XENON1T configuration variables. For example Xe100 uses `'pmt_locations'` to store PMT positions whereas Xe1T uses `'pmts'` which holds more information about the PMTs including their position. During #296 a workaround was put in plugin.py to deal with both options. In the future both configurations should use the `'pmts'` variable.

Maybe related to #222

JelleAalbers:
Fixed in #339.

PR #339, opened 2016-04-04
Consistent pmt configuration settings

JelleAalbers:
This puts the pmt locations in the `pmts` dictionary in the configuration, for both XENON100 and XENON1T -- see issue #300. The rule is now:
  * "Static" options -- anything we don't need to calibrate and that requires a hardware modifiction to change -- are in the pmts dictionary.
  * Other options (such as gains, QEs, etc) have separate configuration options. This makes it easy to change them in different configuration files (e.g. set gain to 1 for LED mode). 

This also sets the assumed gain to 1 in XENON100_LED and XENON1T_LED. That implies areas derived with these configurations will be in electrons (streaming out of the PMT anode), not photoelectrons.


JelleAalbers:
@tunnell Any objection to this?

PR #341, opened 2016-04-08
G4 Optical - FAX interface updated. Fix one bug about the simulated p…

mcfatelin:
Fabio reported the error showed up when trying to use fax-g4 optical interface to process 5000-event MC file. The major error message 

  File "/home/fabiovalerio/anaconda3/envs/pax/lib/python3.4/site-packages/pax/simulation.py", line 270, in make_pax_event
    righter_index, len(current_wave)))

I went to investigate that and found a bug in the fax. In short, the PMT after pulses are generated after the simulated event's time window is defined by the photons produced by S1, S2 and S2 after pulses. Therefore if the PMT after pulse is out of the right boundary of the event window, it raises an error message. This happens very rarely, but once it happens it stops the process. So for simulating large amount of waveforms, this is a bug need to be fixed. 

For fixing it, I simply abandoned the PMT after pulses if it is out of the event window.

### DIFF
--- a/pax/simulation.py
+++ b/pax/simulation.py
@@ -263,9 +263,11 @@ class Simulator(object):
                             len(generated_pulse), left_index, righter_index))
                 elif left_index < 0:
                     self.log.debug("Invalid left index %s: can't be negative" % left_index)
+                    continue
                 elif righter_index >= len(current_wave):
                     self.log.debug("Invalid right index %s: can't be longer than length of wave (%s)!" % (
                         right_index, len(current_wave)))
+                    continue
 
                 current_wave[left_index: righter_index] += generated_pulse
 
--- a/pax/simulation.py
+++ b/pax/simulation.py
@@ -258,11 +258,14 @@ class Simulator(object):
 
                 # Abandon the pulse if it goes the left/right boundaries
                 if len(generated_pulse) != righter_index - left_index:
-                    continue
+                    raise RuntimeError(
+                        "Generated pulse is %s samples long, can't be inserted between %s and %s" % (
+                            len(generated_pulse), left_index, righter_index))
                 elif left_index < 0:
-                    continue
+                    self.log.debug("Invalid left index %s: can't be negative" % left_index)
                 elif righter_index >= len(current_wave):
-                    continue
+                    self.log.debug("Invalid right index %s: can't be longer than length of wave (%s)!" % (
+                        right_index, len(current_wave)))
 
                 current_wave[left_index: righter_index] += generated_pulse
 
--- a/pax/simulation.py
+++ b/pax/simulation.py
@@ -264,9 +264,9 @@ class Simulator(object):
                 elif left_index < 0:
                     self.log.debug("Invalid left index %s: can't be negative" % left_index)
                     continue
-                elif right_index >= len(current_wave):
+                elif righter_index >= len(current_wave):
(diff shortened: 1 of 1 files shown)


JelleAalbers:
Hi Qing, well spotted, I did not consider this before. Skipping pulses outside the range sounds like a good approach. 
* The first error message (if `len(generated_pulse) != righter_index - left_index`) is of a different kind, it checks if the generated pulse has the right length and should never be triggered unless there is a bug in pax. Maybe you could keep this error message?
* For the other two, maybe you could change the `raise RuntimeError(...)` to `self.log.debug(...)` instead; then we can still know the pulses are skipped if we do `--log debug`.

pax/simulation.py
```diff
@@ -256,18 +256,16 @@ def make_pax_event(self):
...
-                    raise RuntimeError("Invalid left index %s: can't be negative!" % left_index)
-
-                if righter_index >= len(current_wave):
-                    raise RuntimeError("Invalid right index %s: can't be longer than length of wave (%s)!" % (
-                        righter_index, len(current_wave)))
+                elif left_index < 0:
+                    self.log.debug("Invalid left index %s: can't be negative" % left_index)
+                elif righter_index >= len(current_wave):
+                    self.log.debug("Invalid right index %s: can't be longer than length of wave (%s)!" % (
+                        right_index, len(current_wave)))
```
tunnell: __[flake8]__
* [F821] undefined name 'right_index'

pax/simulation.py
```diff
@@ -256,18 +256,18 @@ def make_pax_event(self):
...
-
-                if righter_index >= len(current_wave):
-                    raise RuntimeError("Invalid right index %s: can't be longer than length of wave (%s)!" % (
-                        righter_index, len(current_wave)))
+                elif left_index < 0:
+                    self.log.debug("Invalid left index %s: can't be negative" % left_index)
+                    continue
+                elif righter_index >= len(current_wave):
+                    self.log.debug("Invalid right index %s: can't be longer than length of wave (%s)!" % (
+                        right_index, len(current_wave)))
```
tunnell: __[flake8]__
* [F821] undefined name 'right_index'

mcfatelin:
@JelleAalbers Yes, I agree. That makes sense, I modified fax accordingly.

pax/simulation.py
```diff
@@ -256,18 +256,18 @@ def make_pax_event(self):
...
-                if left_index < 0:
-                    raise RuntimeError("Invalid left index %s: can't be negative!" % left_index)
-
-                if righter_index >= len(current_wave):
-                    raise RuntimeError("Invalid right index %s: can't be longer than length of wave (%s)!" % (
-                        righter_index, len(current_wave)))
+                elif left_index < 0:
+                    self.log.debug("Invalid left index %s: can't be negative" % left_index)
+                    continue
+                elif right_index >= len(current_wave):
```
tunnell: __[flake8]__
* [F821] undefined name 'right_index'

pax/simulation.py
```diff
@@ -256,18 +256,18 @@ def make_pax_event(self):
...
+                    self.log.debug("Invalid right index %s: can't be longer than length of wave (%s)!" % (
+                        right_index, len(current_wave)))
```
tunnell: __[flake8]__
* [F821] undefined name 'right_index'

JelleAalbers:
Thanks Qing!

PR #342, opened 2016-04-12
Confidence region calculation improvement

pelssers:
This PR implements some further improvements on the confidence contour calculation by TopPatternFit and the underlying PatternFitter.

Improvements:
* Different calculation of the confidence contours.
Contours that go beyond the edge of the TPC were previously not used. Now these partial contours are also used.

* Different storage of the confidence contours. Store the bounding box of the contour in a ConfidenceTuple; The width (dx), height (dy) and one corner point (x0,y0, so the lower left point of the box).

* Added the correct steps on the fit function for 1, 2 and 3 sigma level contours.
Right tailed probabilities for a chi square distribution with 3 degrees of freedom and p = [0.315, 0.05, 0.01]

* Added another statistic to PatternFitter
'Likelihood_poisson', a likelihood function.
L = Product_over_PMTs( Poisson(area_observed, area_expected)). This simple Likelihood works as well or better as the chi2gamma statistic
Calculate -2ln(L) so the same contour calculation can be used.

* Some tweaks to the diagnostic plotting
* Fixed a bug in chi2 statistic


JelleAalbers:
Thanks for this change Bart, the Poisson likelihood statistic sounds very promising. I looked through the code quickly, it looks fine (though I didn't try to understand the details of the confidence level code). 

  * Maybe you could add a posrectest plot showing that the poisson likelihood statistic works better / as good as chi2gamma?
  * You set the `s2_patterns_zoom_factor` to 2. What impact does this have on the speed? If it makes it much slower, might we be willing to live with the less accurate confidence levels? Or could we decrease the grid size to  compensate?
  * Perhaps it would be a good idea to add to the documentation for the various posrec goodness of fit parameters (including the s1 pattern likelihood in the interaction object) that the value is a -2 ln(L) if the likelihood statistic is activated, and an unreduced chi2 otherwise. This will be useful once we start designing cuts on it.

pelssers:
Thanks for looking through the code. The confidence contour computation is not the most clear code but it is actually what happens inside matplotlib's contour function. Tomorrow I'll show (during the CM lightning talk session) that the resulting confidence contours do have perfect coverage.

However they only have this property when setting `s2_patterns_zoom_factor` to 2. With no zoom the maps are too coarse to properly trace contours.

I have not tested the speed impact yet, however since 2*2 more gridpoints are computed a speed reduction factor of 4 is expected. Maybe some code optimization is still possible.

Reducing the grid size is difficult. Especially since at low energy (150pe) a 3sigma contour needs this grid size. At higher energy or lower significance the grid size could be reduced.

I'm now working on the comparison plot between chi2gamma and the Poisson likelihood.

pelssers:
Hi,
As promised a plot showing that the poisson likelihood is (slightly) better than chi2gamma in terms of the error between the true and reconstructed position.

posrec_resolution_compare_1300pe

Regarding the computation of the confidence contours:
The speed reduction due to the larger zoom factor is roughly 33 percent (Using posrectest I'm processing 210 events/s with zoomfactor=1, 140 events/s with zoomfactor=2).

With `s2_patterns_zoom_factor = 2` the poisson likelihood gives almost perfect coverage across most of the TPC. With the chi2gamma reconstruction (and the same zoom factor) the coverage is worse.

For `statistic = "likelihood_poisson"`
cov1t_z2_1300pe_100k


For `statistic = "chi2gamma"`
cov1t_z2_c2g_1300pe_100k

This motivates the use of `likelihood_poisson` instead of `chi2gamma`. Regarding the contours `s2_pattern_zoom_factor = 2` has to be used. If this is an issue in terms of speed we can consider not calculating the 3sigma confidence level and shrinking the grid size a little. Another option is to make the grid size energy-dependent (larger for low energy, bigger for high energy).

tunnell:
I don't really understand why the poisson does better.  I thought chi2gamma was building a chi2 from a poisson likelihood. Maybe worth a talk at meeting?  Also, can you vary your bounding box to get a more uniform coverage in radius?

pelssers:
I read through the original article by [Mighell](http://iopscience.iop.org/article/10.1086/307253/pdf) again. This article was the motivation to use the chi2gamma statistic in Xerawdp. It's main conclusion is that in the case of Poisson distributed data one should use this new chi2gamma statistic instead of Pearson's or Neyman's (modified) chi2. The paper also says the chi2gamma statistic is similar in performance to the maximum likelihood ratio statistic for Poisson distributed data. An important difference is that the chi2gamma is inherently weighted and thus easier to include uncertainties (QE, gain, etc.)

I'll repeat the plots now that we have the correct QE in PAX and also run some robustness checks to see how the algorithms deal with partially random hitpatterns.

On your request to vary the bounding box:
I may have misunderstood the question but the coverage (for the likelihood case) looks very uniform to me. The effects at large radii occur because the contour is touching the edge of the TPC. Increasing the bounding box would only shift the coverage not change its shape as a function of the radius.

pelssers:
To test the robustness of the reconstruction I ran PosrecTest on partially random hit patterns. (Simulating a hit pattern, then taking a certain fraction of the hits and redistributing them randomly to simulate a non-ideal still). Using 10% random hit patterns the likelihood statistic still outperforms the chi2gamma statistic as can be seen in the figure. The difference between the two is actually larger than without randomization.

These reconstructions also use the configuration with the QEs (instead of the placeholder values we used before).

posrec_resolution_compare_r10_1300pe

How do we move forward with this PR? Without it the confidence contours are meaningless.

JelleAalbers:
TopPatternFit is now dominating the speed again:

    +---------------------------------+-------+-------------+------+-----------+
    | Plugin                          |     % | /event (ms) |  #/s | Total (s) |
    +---------------------------------+-------+-------------+------+-----------+
    | ReadZipped                      |   0.1 |         0.4 |      |       0.0 |
    | DecodeZBSON                     |   4.3 |        12.0 | 83.4 |       1.2 |
    | SortPulses                      |   0.1 |         0.4 |      |       0.0 |
    | ConcatenateAdjacentPulses       |   0.1 |         0.3 |      |       0.0 |
    | CheckBounds                     |   0.2 |         0.4 |      |       0.0 |
    | FindHits                        |   3.6 |        10.0 | 99.5 |       1.0 |
    | GapSizeClustering               |   1.3 |         3.7 |      |       0.4 |
    | RejectNoiseHits                 |   1.5 |         4.1 |      |       0.4 |
    | SumWaveform                     |   4.3 |        11.8 | 85.0 |       1.2 |
    | BasicProperties                 |   6.8 |        18.9 | 53.0 |       1.9 |
    | SumWaveformProperties           |   8.3 |        22.8 | 43.8 |       2.3 |
    | CountCoincidentNoisePulses      |   0.0 |         0.1 |      |       0.0 |
    | PosRecWeightedSum               |   0.7 |         1.8 |      |       0.2 |
    | PosRecMaxPMT                    |   0.3 |         0.8 |      |       0.1 |
    | PosRecRobustWeightedMean        |   1.2 |         3.3 |      |       0.3 |
    | PosRecNeuralNet                 |   1.5 |         4.2 |      |       0.4 |
    | PosRecTopPatternFit             |  41.7 |       115.2 |  8.7 |      11.6 |
    | HitpatternSpread                |   5.0 |        13.7 | 73.0 |       1.4 |
    | GasXenonZeroFieldClassification |   0.0 |         0.1 |      |       0.0 |
    | BuildInteractions               |   0.0 |         0.1 |      |       0.0 |
    | BasicInteractionProperties      |   0.0 |         0.0 |      |       0.0 |
    | DeleteLowLevelInfo              |   0.5 |         1.3 |      |       0.1 |
    | EncodeROOTClass                 |  16.4 |        45.3 | 22.1 |       4.6 |
    | WriteROOTClass                  |   1.9 |         5.2 |      |       0.5 |
    | TOTAL                           | 100.0 |       276.0 |  3.6 |      27.9 |
    +---------------------------------+-------+-------------+------+-----------+

It's not a big issue now, but will likely be once we have events with many more peaks. Maybe we can try dropping the 3sigma contour and shrinking the gridsize, as you suggested?

pelssers:
Hi Jelle,
Shrinking the grid size is the best way to speed up the plugin. However there is also an energy dependency involved since low energy S2s need a larger grid size. 
Currently the Xe1T configuration uses a very large 10*10 cm grid. A typical 150pe S2 results in these contours:

plot_0 37_40 27

I'll try and optimize these settings for speed. Going from a 10*10 cm grid to 7*7 cm should double the processing speed but means losing at least the 3sigma contour.
I'll see if there is a way to vary the grid size dynamically

tunnell:
Shouldn't you wait until we get S2s before we think too much about optimizing this?

pelssers:
If you can live with the current processing speed then it makes sense to optimize later. If not I can you give you a factor 2 speed increase with the 7x7 grid now and optimize later. It's only one commit difference.

JelleAalbers:
I like speed :-) Factor two would make me happy if it's just a small tweak.

pelssers:
Done, can you run the same test as you did before to see the actual speed increase this gives?

Issue #343, opened 2016-04-13
Trigger data should pass through MongoDB


JelleAalbers:
Currently the trigger data is just dumped to an hdf5 file. This is not readable while the run is ongoing.

We should make a separate trigger data database in mongo, with a collection for each run, where this data can temporarily reside. After the run has ended the contents should be dumped to a file, so the analyst can find it along with the ordinary raw data (zipped bson). The trigger data can then be deleted from Mongo (if it takes up a lot of space, otherwise we can just keep it a while so people can continue to browse).

JelleAalbers:
Implemented in #347

PR #347, opened 2016-04-24
Trigger monitoring data to MongoDB

JelleAalbers:
This changes the data format of the trigger monitor data from HDF5 to MongoDB (see #343) + a zipfile in the raw data directory. This should enable monitoring of the dark count data while the run is ongoing.

For Mongo, it makes a collection on the same event builder where the untriggered data is stored, but in the 'trigger_monitor' database. Name of the collection is again name of the run.

The format in both Mongo and the zipfile is somewhat arcane, but here is a notebook showing you how to load it into a less arcane format and make some plots, either MongoDB or the zipfile:
https://github.com/XENON1T/XeAnalysisScripts/blob/master/examples/Trigger_monitor_analysis.ipynb

This also adds a `--stop_after_sec` option for the event builder to build only part of a run.


coderdj:
Hi Jelle,

Why not use MongoDB's document format for storing the data? MongoDB docs are basically dictionaries but they have the nice feature that they're easy to dump to JSON. But what you do is serialize numpy objects into a single field in mongo. If this is meant to be web-friendly it might be better to avoid python-specific classes.

tunnell:
Or bcolz

JelleAalbers:
Superseded by #348

PR #348, opened 2016-04-28
Parallel trigger queries + Monitoring data in MongoDB

JelleAalbers:
This includes

* Allow the trigger to execute **several queries for pulse times in parallel** (using a ProcessPoolExecutor). This significantly increases performance (essentially linearly with the number of parallel queries) at the expense of additional CPU load. The maximum number of workers is controlled by the `max_query_workers` in the `[Trigger]` section of the configuration. Each query always asks for one `--batch_window` of data; if there is not enough data to query, less workers will be used.  
* A `--delete_data` option, which lets us **delete data after triggering** it. The deletes are done just after the read queries, so they are also done in parallel: this is necessary because the delete performance seems much worse than the read query performance. The deletes are done via unacknowledged writes (writeconcern `w=0`, journalling is already disabled completely in our mongo deployment settings). Note this is only useful for live triggering: when a run is already done, it is much more efficient to trigger without deleting, then let cax drop the collection when everything is well.
* The trigger's **query for pulse times is optimized** to use only the index, and no longer count the documents first -- using monary's `block_query` instead of `query`. Currently the digitizer module and channel number are not yet in the index, so per-PMT dark rate monitoring will be nonfunctional until we fix this.
* The trigger's **monitor data is now saved in MongoDB** as well as a zip file. For more details see the pull request #347, which this PR supersedes. This also includes a fix for a bug that prevented folders with with `trigger_monitor_data.zip` in them to be read for processing.


tunnell:
Once it passes tests, it's fine to merge.

PR #349, opened 2016-04-29
Minor XENON1T config fix

DCichon:
Fixed some signal/HV connector numbers being wrong in the XENON1T config file (example: ’3.10’ was saved as ’3.1’).


DCichon:
Wait with pulling, as there is the same issue if the signal/HV connector channel number is 20 (and possibly other multiples of 10).

JelleAalbers:
Thanks Dominic, good to remove this error. It would be even nicer to store the connector and channel numbers in separate fields as integers though.

PR #354, opened 2016-05-04
Fixed a case of the FaX physics backend trying to use a non-existing logger.

DCichon:
A small fix for a bug I encountered while playing around with FaX (if indeed the simulation module logger is to be used).


JelleAalbers:
Thanks, one less bug!

PR #355, opened 2016-05-04
Implemented a new 'LED' recoil type.

DCichon:
Photons are emitted uniformly within a configurable time interval instead of following excimer decay time distributions, resulting in an LED-like light source.

This addition is motivated by making PMT afterpulse measurement simulations possible with FaX. If needed, a better LED simulation could be achieved in the future by dedicated measurements of the LED hit time distribution ([similar to Jelle's hitfinder calibration note][1]).

[1]: https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon100:analysis:led_pax#when_does_the_led_fire


tunnell:
Payam, can you check these values before we merge?

JelleAalbers:
Payam reviewed this (see the pax gitter). There is an issue with the travis build which is fixed in master, so I'll merge this.

PR #357, opened 2016-05-06
Update XENON1T_LED.ini

pdeperio:
Request from Payam for LED tests this weekend.

### DIFF
--- a/pax/config/XENON1T_LED.ini
+++ b/pax/config/XENON1T_LED.ini
@@ -14,6 +14,8 @@ pre_output = []
 # [HitFinder.FindHits]
 # make_diagnostic_plots = 'always'
 
+[HitFinder]
+height_over_noise_high_threshold = 3
 
 [DEFAULT]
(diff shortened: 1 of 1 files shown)


pdeperio:
```The command "python tests/test_maxpmt.py" failed and exited with 1 during.```

@JelleAalbers @Payampakarha

JelleAalbers:
Hi Patrick, this error (or rather the error that underlies it, this message just indicates none of the tests even want to start) currently affects all our travis builds. Most likely one of our dependencies released a new version which causes some problem, at least on the travis server. @cwittweg saw this error too and resolved it by installing an earlier version of scipy. I will try to trace down the problem.

Someone reading this might suggest we should pin/freeze all our dependencies to specific versions to avoid these issues from occurring again. Appealing as this sounds, we would miss out on new features, speedups, but most importantly bugfixes. We won't even be aware of a bug if we pin the version and leave it there. But let's see how often these things happen.

pdeperio:
is it safe to merge then?

PR #360, opened 2016-05-18
Use datasheet gains instead of 2e6

JelleAalbers:
This makes pax use the Hamamatsu datasheet gains rather than a uniform 2e6 gain. We've seen in recent studies (https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:aalbers:gains_from_datasheets_and_lone_hits, https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:saldanha:xe1t:prelim_gain_cal) the gains are very far from uniform and very far from 2e6 (more like 5e6), but the datasheets gains are quite reasonable. PMTs that are known to be silent are 'turned off' in software (gain=0). 

While we discuss a better set of gains to use, these will give more reasonable results with the S2 data we get in the coming days. The gains are stored in the root file, so we can always recover which gains were used for which processing.


Payampakarha:
Hi Jelle,
This is nice to do. By today and tomorrow, I can also provide a list of reliable gain values at 1500 for some ~100 PMTs. Maybe we can replace the gains of these list right away from the datasheet to the calibrated values.

tunnell:
Don't merge this until we discuss Monday.

JelleAalbers:
Then we can close this, since on Monday we can probably decide better gains. Don't you want reasonable gains for this week's data?

tunnell:
We can use datasheet for problematic PMTs.  I'd leave this open.  Maybe we want to use this until @Payampakarha has official gains since we show the datasheet ones aren't that bad.  It's a group decision.

JelleAalbers:
Superseded by #361

PR #361, opened 2016-05-18
Preliminary settings for XENON1T processing

JelleAalbers:
Now that we have the first S1 + S2 dataset, we can try to use some preliminary settings to do sensible processing. This pull request does the following:
 * Activate the Hamamatsu datasheet gains (superseding #360)
 * Make the GapSize clustering split a peak whenever there is no hit for 100ns. This likely splits up single electron S2s, but is much better for high energy events.
 * Active the natural break clustering with the XENON100 settings (previously it was turned off).
 * Clone the AdHocClassification plugin from XENON100 to XENON1T, and tweaks a few settings:
  * Everything below 50 pe area is marked as 'unknown'
  * I slightly tweaked the S1/S2 width boundaries.
 * Calculate the drift time from the maxima of the S1 and S2 peaks, rather than their centers of gravity. This is essential for high energy (and caused an artifact I discuss at the end of my [electron lifetime example](https://xecluster.lngs.infn.it/dokuwiki/lib/exe/fetch.php?media=xenon:xenon1t:aalbers:electron_lifetime_analysis_example.html)).
 * Fix the time displayed in the event plots to be the actual UTC time.

To test these I processed 1000 events from run 569 (160518_1342), the first S1+S2 dataset. Hopefully we can process a few complete datasets with these settings in the coming days, so we can do the next iteration of settings in the next analysis telecon.

Below you see the area-width histogram of all non-lone-hit peaks, with the classification lines indicated. Everything left of the vertical line and in the middle box right of the line is 'unknown', everything in the top right box is 's2', the bottom right box 's1'.

class_bounds

As you can see, the classification bounds are reasonable for a first processing. When we have more statistics we will know better. There is a lot of junk at low energy, probably because of too gregarious clustering settings, and because we have a high rate of small thingies anyway in these conditions.

As a further test, here are the (x, y) position distribution of the events with an S1+S2 pair, as reconstructed by the TopPatternFit algorithm on the S2:

tpf_posrec

This shows a clear excess at where we expect the Cs source to be. 

_Text and plot removed by C. Tunnell 19 May since XENON internal._


sanderbreur:
Very nice update and great to see the tools working.

If we can merge this asap we can start reprocessing and thus analyzing the
S1 and S2 events properly.

Cheers,
S

Op wo 18 mei 2016 om 19:40 schreef Jelle Aalbers <notifications@github.com>:

tunnell:
You can already reprocess with this without it being master yet.  It would be a useful step for checking this and this work + check are worth discussing Monday.  I think we'll merge before Monday then just update everybody.  

Sander, are you willing to look at this?

sanderbreur:
Yes, no problem, will do it tomorrow first thing.

S

sanderbreur:
Ok, I have done the following:
- Checked the code for typo's
- Checked if the PMT list of gains is correct with comparison to Richards findings
- Checked the new peak classification in area vs width space

@tunnell I see no problems with this pull request and I would advise to merge this asap. 

Improvements to be made after we process and analyse more data with this pull request can be discussed Monday during the analysis telecon and should include:
- Discuss where we would like our S1 and S2 classification to end. (right now some of out gaseous S1's are being classified as 'unknown') This is a subject someone should spend a few weeks on to improve the method and fine tune it.
- Waveform watching to find out where our clustering works and where not. (Someone should work on the high energy clustering for a few weeks)
- Make an open issue for someone to investigate the dt calculation (also to make sure the waveform sim and data can be compared correctly)
- Improvements we should now see that 'first order' gains have been implemented. 
- How or when to implement the S1 and S2 LCE maps

tunnell:
I'll merge and make minor release after builds?

PR #363, opened 2016-05-20
Use highest baseline from start and end of pulse

JelleAalbers:
This makes the hitfinder compute the baseline at the start AND the end of each pulse. The highest of the two is used as the reference point (and stored in pulse.baseline), the difference is stored in the new attribute pulse.baseline_difference. 

First, this makes baseline shifts detectable in the output. Second, it prevents overshoots (probably caused by the amplifier) after large signals to become part of "endless" hits that last until the end of the pulse. In a sense this is a similar problem to the one we had in the DAQ firmware. 

| Before | After |
| ------------- | ------------- |
|  event0032_pulse05054-05341_ch139 | event0032_pulse05054-05341_ch139 |
| event0032_pulse05056-05569_ch214 | event0032_pulse05056-05569_ch214 |
| event0040_pulse05039-05352_ch033 | event0040_pulse05039-05352_ch033 |
| event0040_pulse05042-05563_ch139 | event0040_pulse05042-05563_ch139 |

(events are from run 588)


JelleAalbers:
The baseline increase is now also written in the hitfinder diagnostic plots:

event0093_pulse05085-05340_ch139



There is a new possible value for `make_diagnostic_plots`, 'baseline shifts', which plots all pulses which show more than 10 ADCc baseline shifts.

It looks like the shifts are predominantly (but not exclusively) occurring in a few channels, such as 33, 139, 214. Interesting...

tunnell:
It seems like a simple solution that should work fine.  Are there any more general solutions you considered since detrending is a quite common thing with electronic engineering.

We need to be quite careful with this since the baseline difference needs to be incorporated into the hit area uncertainty.  This will become more important for few pe S1s.  If we don't check this difference in the analysis, bad things can happen.  How should we flag crazy values either within pax or at analysis stage?  

Does this require reprocessing all data?

tunnell:
Closed and superceded by #369.

PR #364, opened 2016-05-25
Fixes/additions to PMT map in ini

JelleAalbers:
These are some small changes to the cable map json in XENON1T.ini:
* Separate connector and channel fields. So no more strings '8.04', but separate integer fields `connector=8` and `channel=4`.
* PMT serial numbers now stored as ints, not floats 
* Add amplifier information. I thought I already did this... but apparently I forgot to commit it.

I also correct an error I introduced in #361: the placeholder gain for the diagnostic PMTs should be 2e6 rather than 4.8e6, since these are 1-inch XENON100 PMTs running at similar voltages as in XENON100.


tunnell:
ok

PR #365, opened 2016-05-25
Saner trigger defaults

JelleAalbers:
This sets the default trigger settings to the ones we used over the last few weeks: trigger always above 50 pulses, with 50 us left and right of each trigger. This will capture high-energy events up to 100us apart (since both the S1 and S2 trigger).

Last week we set these settings manually or via a custom config file. When the rate has gone down and the lifetime has gone up, we should of course put wider left and right extensions.


JelleAalbers:
If there are no objections I will merge this soon, since it ensures continuity of trigger settings if we take data tonight.

tunnell:
not sure when we take data...

Issue #65, opened 2014-10-20
INI files autocreated from run DB


tunnell:
Maybe these should be automatically created?  That way we can run offline, but we can manage all the info in a more sane way with a DB.  Also, we have too much documentation in the ini files that should move to the docs.

EDIT: This will include calibration constants, electron lifetime, etc

JelleAalbers:
Since we can now specify arbitrary configuration for pax in the run doc, can we close this?

PR #366, opened 2016-06-01
Configuration settings from runs db

JelleAalbers:
This enables pax to read configuration instructions from the runs database (see #65). The 'processor' section of the run doc is treated as a config_dict for pax. For example, if you set

    'processor': {
        'DEFAULT': {
            'gains': [7, 7, 7, ..., 7, 7],
        }
    }

it will use the gain 7 for all PMTs.

How it works 
-------------------
When the option `look_for_config_in_runs_db` is set in the [pax] section, it will try to find the run number for the run you are processing (and error if it cannot). This is set for XENON1T by default; Simulation.ini turns it off explicitly.

Pax now tries to find the run document (using the public Stockholm mirror for the runs db):
  * If the `run_number` option is set to a nonzero value, it will use that. When you make your own Processor class (as cax does for example) this should be the preferred method.
  * Else, it will determine the run name from os.path.basename() of the input_name option (the thing you pass with --input). This is only for convenience use with paxer.

Next, the configuration will be reloaded with the runs db settings merged in. The run database settings have higher priority than the configuration files (even the user-specified ones), and even the config_str you can pass on the command line. The config_dict does have higher priority than the runs db. 

If you want to bypass the runs db settings, you can set `look_for_config_in_runs_db` to False. If you want to override just one runs db setting, use config_dict (or if you really can't, extract the other settings from the runs db to a config file). 

Other changes 
-------------------

* The ClientMaker (convenience for MongoDB connections) is moved to a separate file. This avoids having to do an import from a plugin in the core.
* The MongoDB connection settings have moved from hardcoded defaults in event-builder to XENON1T.ini. eventbuilder.ini overrides the options that are different for the event builder (hostname, user).
* I moved the ZLE/self-trigger emulation settings to TPC-specific config files, since they are different for XENON1T and XENON100.
* Various configuration errors in core.py now got their own exception class InvalidConfigurationError. If a mongo connection was made succesfully, but no run doc matching the given run_number / input_name is found, this also raises InvalidConfigurationError.
* Added hide_peak_info and hide_trigger_info options for plotting. These hide the peak labels / ranges and trigger signal markers from he plot. Useful for producing public plots. (sorry for tagging this along, may have been better as a separate PR).


tunnell:
Looks good.

PR #367, opened 2016-06-01
Avoid numpy runtime warnings from failed confidence contours

JelleAalbers:
We currently get the message

    /home/jelle/anaconda3/envs/pax/lib/python3.4/site-packages/numpy/lib/nanfunctions.py:227: RuntimeWarning: All-NaN axis encountered
      warnings.warn("All-NaN axis encountered", RuntimeWarning)

every ~50 events so. This happens when a list in the confidence contour calculation contains only NaN values. 

@pelssers maybe you'd like to have a look at these edge cases later. For now I just avoided the warning by leaving the confidencetuple's values at nan (which would probably have been the result anyway).


pelssers:
Understanding these contours (mostly at very large radii) is something I'm working on, but I agree that this (harmless) numpy warning may as well be suppressed. Thanks for implementing!

Issue #225, opened 2015-09-01
Dealing with high-impurity datasets


JelleAalbers:
The cesium dataset we picked today (xe100_ 111109, see also #221) is coincidentally one from just after the chiller problem in run 10, in which the impurity concentration is very high:

https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon100:analysis:electron_lifetime_fit_with_polynomials

This causes the electron train at the tails after S2s to be much stronger. Even with modestly large S2s (below 100k pe), this causes the clustering to fail utterly:

![000006](https://cloud.githubusercontent.com/assets/4354311/9614955/67be8dba-50f5-11e5-8402-9aabf2be4260.png)

For comparison, here is an AmBe event with a more than 5x as large S2, which we can deal with:

![000000](https://cloud.githubusercontent.com/assets/4354311/9614964/7e2c7418-50f5-11e5-9ae2-ae1807e01b4a.png)

We are probably going to encounter conditions like these in XENON1T commissioning, so we should at least include options to deal with these conditions.

JelleAalbers:
The drift time is now calculated from between the maxima of peaks, which should be unaffected by 'sticky tails'. With the new clustering (#369) I hope the S2 width also becomes more reliable in high-impurity conditions.

JelleAalbers:
Adressed by #369 and possible future re-tuning of that.

PR #369, opened 2016-06-01
Local-minimum clustering, hitfinder tuning

JelleAalbers:
This includes a new clustering plugin aimed at high-energy/high-rate conditions (see #225), and a second round of basic tuning for the XENON1T hitfinding, clustering and classification (see #361 for the first round).

Local-minimum clustering
----------------------------------
This includes a new clustering plugin, run before the (unchanged) natural breaks clustering, which splits large peaks at local minima in their sum waveform. The peaks on both side of the minimum have to be a configurable multiple of the minimum, and a configurable absolute length above the minimum. These settings avoid splitting signals which naturally have a minor double-peaked feature (shallow S2s) or pairs of S2s which are very different in size, for which the split would yield nonsensical areas.

We could add a waveform filter in front of this clustering, as is done in Xerawdp, then see if we can lower the `min_height` and `min_ratio` values. However, this is mostly beneficial for low- and medium energy signals, for which the natural breaks clustering seems to work quite well (at least for XENON100).

I set the GapSizeClustering threshold from 100 to 250 ns, which would split up less than half the single electrons (see figure 3A in [this note](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:processor:natural_breaks_clustering#pax_s_clustering)) if the single electrons looked the same as in XENON100 (they probably don't). We decreased this value earlier to improve the separation efficiency for high-energy signals (#361), but LocalMinimumClustering can do this far better.

Below are some example waveforms before and after this PR. Look at the peak labels in the lower panel: you clearly see the high-energy peaks are better separated. There is also some improvement in the fragmentation of the low-energy tail after the S2 (due to the GapSizeClustering setting change).

| Before | After |
| ------------- | ------------- |
| 000000_old |   |
| 000130_old |  |

Hitfinder tuning
--------------------
First, this includes the fixes from #363. We can close that PR, unless it's useful to compare those changes separately.

Second this tweaks some hitfinder settings:
  * low_threshold from 1 to 2 sigma. This tightens the integration threshold around each hit, which will probably increase the bias in our hit areas in return for less variance.
  * dynamic_low_threshold_coeff from 0.01 to 0.1. This raises the low threshold (and therefore effectively also the high threshold) of the hitfinder after a hit for the remainder of the pulse of 10% of the hit height. 

Both of these help avoid long hits due to the (over|under)shoots probably due to the amplifier (see also #363), and the first also avoids including a lot of noise in the integration.

| Before | After |
| ------------- | ------------- |
|  event0005_pulse00350-00453_ch038 | event0005_pulse00350-00453_ch038 |
| event0000_pulse05066-08173_ch108 | event0000_pulse05066-08173_ch108 |

Classification recalibration 
-----------------------------------
With the new clustering settings, the classification also has to change a bit. The big blob of stuff at low-area, high width (mostly coincidences of dark rate) has receded from ~50pe to 10pe, revealing a population around 20pe which might be the single electrons (but don't bet on it yet). Mostly this is due to the GapSizeClustering setting.

| Histogram  | Classification |
| ------------- | ------------- |
| areawidth_histogram  | areawidth_classification  |

* For comparison with the old classification, see Sander's plots under 'peak level checks' [here](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:analysis:first_cs_s1s2_checks#peak-level_checks). 
* The plots show all 646403 non-lone-hit tpc peaks from the first 10k events from run 599 (a Cs dataset with anode at 4kV). 
* I labeled the population which might be single electrons, and the population of pile-up of multiple of these, as s2s, as we did for XENON100. We should consider whether this is a good idea later.
* The gas S1s are still labeled as S1s.
* The classification is not perfect, in particular, some of the very high width dark rate pile-up peaks are labeled as very small S2s (below 20pe or so) rather than unknowns. This is unlikely to be confusing in the analysis, but could be a nice exercise to improve.

Tests and tuning
-----------------------
Unfortunately we can't optimize this on simulated data yet: we have to tune the waveform simulator to the XENON1T data first, which probably requires quieter conditions (in particular, to isolate single electrons better). Hence, the only way to test and tune it is to look at a lot of waveforms (also PMT waveforms). I've selected some for the 'before-after' plots above, but of course looked at more of them.


pax/plugins/peak_processing/ClassifyPeaks.py
```diff
@@ -10,17 +10,21 @@ def transform_event(self, event):
...
-                    peak.type = 's2'
-            else:
+            if width > 0.1 * area**3:
                 peak.type = 'unknown'
 
+            elif width < 50:
+                peak.type = 's1'
+
+            else:
+                if width > 3.5e2 / area**0.1 or width > 1.5 * area:
```
tunnell: This is just some fit to the data?  Just curious where these values come from.

pax/plugins/peak_processing/LocalMinimumClustering.py
```diff
@@ -0,0 +1,131 @@
...
+            if not len(hs):
+                raise RuntimeError("Attempt to create a peak without hits in LocalMinimumClustering!")
+
+            yield datastructure.Peak(left=l,
+                                     right=r,
+                                     detector=peak.detector, hits=hs)
+
+
+#  @numba.jit(numba.float64[:])
+# TODO: TESTS!
```
tunnell: :)

pax/plugins/peak_processing/LocalMinimumClustering.py
```diff
@@ -0,0 +1,131 @@
...
+    for i, x in enumerate(w):
+        if x < min_since_max:
+            # New minimum since last max
+            min_since_max = x
+            min_since_max_i = i
+
+        if min(last_max, x) > max(min_since_max + min_height,
+                                  min_since_max * min_ratio):
+            # Significant local minimum: tell caller, reset both max and min finder
+            yield min_since_max_i
```
tunnell: So this works on the sum waveform.  It's unfiltered.  And you just find some dip.  I'm just curious if it makes sense to do filtering.  Apparently, it doesn't matter, but this is moving ever so closer to a KDE.

tests/test_hitfinder.py
```diff
@@ -11,7 +11,7 @@
 class TestHitFinder(unittest.TestCase):
 
     def peak_at(self, i, amplitude=10, width=1):
-        w = np.zeros(100)
+        w = np.zeros(200)
```
tunnell: ?

pax/plugins/peak_processing/ClassifyPeaks.py
JelleAalbers: Yes, I just made the histogram shown in the pull request and drew some lines.

tunnell:
The code at least looks fine.  Tiny comments. It's sort of unfortunate that we have to mix two types of clustering: 1D clustering with this Natural Breaks and then this minimum finder.  This method seems to work and is mainly important for the initial data.  It should never operate on our low-energy scatters, right?  I still think we could do something smarter like KDE or something, but this seems fine for the data we're taking in the coming months.  

If @planteg finds the method fine and after he comments, we can merge after his review.

pax/plugins/peak_processing/LocalMinimumClustering.py
JelleAalbers: Yes, I think it makes sense to try at least. For S2 splitting it should work well, for S1 splitting natural breaks should take care of most of the work. Once we know roughly what our S2 widths will be we can find out what filter settings make sense.

tunnell: We already know roughly what they are.  They aren't 10 ns but more like a microsecond.

planteg:
This method of splitting overlapping S2s at the minimum between a pair by requiring the two S2 heights to be a given amount higher than the minimum and a given factor larger than the minimum looks good to me. Naively it seems like this can work well on unfiltered waveforms until the fluctuations in the waveform (from a lower photoelectron rate) are comparable to the minimum height parameter. If the current clustering algorithm works well from the resolved single photoelectron regime (single electron S2s) up to somewhat higher than that regime, which we think it does, then the gains would be minimal by adding filtering.

PR #372, opened 2016-06-06
Increase trigger window

tunnell:
TBD.


tunnell:
@JelleAalbers I guess this is fine for now.  We'll just keep increasing it as detector conditions improve.

PR #380, opened 2016-06-16
Address new conda version

tunnell:


JelleAalbers:
For python 2 it's trying to install many packages from source for some reason?

Issue #359, opened 2016-05-17
Log event builder errors


JelleAalbers:
When the event builder crashes, it does not write the error trace to the event builder log.

JelleAalbers:
The current way we handle the event builder log really does not work; looking through the logs from last night I see several logs partially overwritten by the start of the log from the next run.

JelleAalbers:
Addressed by #381.

JelleAalbers:
We should still not immediately delete the log when the next run starts, the copying out may have failed.

JelleAalbers:
Addressed by #385.

PR #381, opened 2016-06-16
Trigger config from runmodes

JelleAalbers:
This allows us to specify overrides for the trigger configuration in the run doc. For example, have a look at the alphas_stable runmode, which has this:

    "trigger_config_override": {
        "Trigger|DecideTriggers": {
          "trigger_probability": {
            "0": { "2": 0},
            "1": {"50": 0.01, "100": 1},
            "2": {"50": 0.01}
          }
        },
        "Trigger|ClassifySignals": {
          "s1_max_rms": 12
        }
    },

which prescales normal signals by 1%, but lets narrow, large-area S1s pass. I tested this with a small run, it seems to work.

The `|` is necessary because `.` in field names has a special meaning in Mongo. It is translated to `.` when the config is loaded.

Also includes a minor plotting fix, and adresses #359 by writing a trigger exception to the log and marking the trigger status of the run as 'error' (as opposed to leaving it 'transferring' forever).

### DIFF
--- a/bin/event-builder
+++ b/bin/event-builder
@@ -21,7 +21,7 @@ import shutil
 from six.moves import input
 import pymongo
 
-from pax import core, units
+from pax import core, units, configuration
 from pax.MongoDB_ClientMaker import ClientMaker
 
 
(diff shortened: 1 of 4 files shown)


JelleAalbers:
I'm going to merge this, since it includes error reporting for the trigger, and we've been running with this for a while without problems (alphas_stable only works with this). Hope Santa isn't watching.

PR #382, opened 2016-06-17
XENON1T S1 relative light yield map

JelleAalbers:
This adds the Monte Carlo prediction for the XENON1T S1(x,y,z) relative light yield map. I used the simulation data from @massoli (the same files as used for the per-PMT map), then checked that the results agree with figure 14 from the Monte Carlo paper (at least roughly; this is an xyz map rather than a rz map, so comparison is a bit interesting anyway).

I've updated the conversion/verification code in https://github.com/XENON1T/XeAnalysisScripts/blob/master/mapmaking/Parse_Fabio_map.ipynb with the necessary changes. I tested it in the same way as we initially did for XENON100: I asked the values of the light yield map directly from pax in the same way the correction code does, and get the right light yield map out:

s1_lce_xyz

As with XENON100, the correction applied by pax to S1s is the inverse of this map. The normalization is chosen so that a peak from a uniformly distributed source will not move due to the correction. This map does not take silent channels into account. Once we've done an internal source calibration, the map can be compared to and replaced with a data-derived map (as it was for XENON100). For more details on the Monte Carlo that was run to create this map, please see https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:fvm:single_pmt_lce_maps.

Some smaller changes:
  * Update the S1(x,y,z) patterns map: the topmost bins in z were identically zero for all PMTs. Probably no big deal (the same is true for positions outside the TPC radius, and it does not confuse the algorithm). 
  * Add a convenience argument `--ignore_rundb` to pax which allows you to process XENON1T data without going to the run db for override settings.


tunnell:
Why not use the absolute LY?

tunnell:
i.e. corrected units of photons

JelleAalbers:
That's possible, it will just mean dividing/multiplying by a constant (which we can measure). It would be different from what we did in XENON100, but more consistent with what we do for the electron lifetime.

Even so, I would favour to store the map as a relative light yield and the absolute light yield as a separate parameter in the config. If the light yield changes, we can just adjust that parameter and don't have to re-make the map.

tunnell:
I agree with splitting it as you say.  Can we put in the LY feature now though?  At least a rough value as placeholder?

JelleAalbers:
Ok, let's add a new interaction attribute s1_absolute_scaling to track this scaling. For a single dataset it is redundant with the configuration metadata, but when combining datasets it will be useful. Then we combine s1_absolute_scaling with s1_spatial_correction and s1_saturation_correction into s1_area_correction.

We should probably do the same for S2: add an absolute scaling attribute to represent the SE gain and extraction efficiency (if it turns out we have any) and add it to the existing corrections. Then we'll report both cS1 and cS2 in units of estimated quanta originally produced in the interaction.

We'll probably get asked for new columns in the Basics hax treemaker for the 'old' way of doing corrections, e.g. for studies which want to determine the absolute light yield.

PR #388, opened 2016-06-21
Config check to prevent invalid event range settings

JelleAalbers:
This raises an exception if somebody tries to use event size settings that can lead to overlapping events. See also https://github.com/XENON1T/pax/commit/54ea7b177a3391cc314ab2484c016d9843c07ab1.


pax/trigger.py
```diff
@@ -75,6 +76,11 @@ def __init__(self, pax_config, trigger_monitor_collection=None):
         self.monitor_cache = []         # Cache of (data_type, doc), with doc document to insert into db / write to zip.
         self.data_type_counter = defaultdict(float)    # Counts how often a document of each data type has been inserted
 
+        # Configuration sanity check
+        if self.config['event_separation'] < self.config['left_extension'] + self.config['right_extension']:
+            raise InvalidConfigurationError("event_separation must not be smaller "
+                                            "than left_extension + right_extension")
```
tunnell: __[flake8]__
* [W291] trailing whitespace

PR #389, opened 2016-06-23
Fax g4 interface update

mcfatelin:
FAX-G4 MC interface updated. 

Main updates:
1) Pseudo simulation mode using MC root file is updated. 
2) Also fax is able to use the same MC root file for optical simulation mode. 
3) Temporary S1 LCE map is added for avoiding out of LCE region error. The temporary LCE values for below cathode is the same as at cathode.

To run now the fax waveform simulation using the MC root file as inputs:
1) For pseudo mode:
paxer --input <MC root file> --config XENON1T SimulationNESTInput .....

Under this mode, fax takes the number of photons and electrons of each event from MC root file, and put other hardware- and detector-related characteristics (PMT shaping, digitizer shaping, LCE and gas amplification etc) into the final pulse shape. 

2) For optical simulation mode:
paxer --input <MC root file> --config XENON1T SimulationOpticalInput .....

Under this mode, fax takes directly the info of photon tracking information from the MC root file, and just apply the PMT and digitizer shaping into the final waveform pulse shape.

Note in both the pseudo and optical full simulations, the after pulses (PMT and photoionization) are generated. The generated after pulses are always the pseudo approaching.


JelleAalbers:
Hi Qing, thanks for updating the interface! I don't see any obstacle to merging as soon as possible.

If the new xyz S1 per pmt-map is simply an extension and update of the previous map, I think we can make it the default in XENON1T.ini. If you agree, maybe you could show a small plot comparing the maps (e.g. an (r, z) LCE for one PMT), and then we can make your map the new default.

mcfatelin:
Hi Jelle. Thanks for the suggestion. I plot the LCE map for PMT 176 on Z verse R, for the previous and my extended pattern map. It is basically a simple extension.

Previous pattern map:

pmt176_lce_previous


Extended pattern map:

pmt176_lce

JelleAalbers:
Thanks for the plot! I will talk with Chris tomorrow to see if we can merge and release (together with some other pull requests). 

As you mentioned in your email, another option to deal with the missing LCE knowledge below the cathode, while we wait for a full LCE map, is to ignore out-of-range events in the NEST output / simulation instructions (or add an option to enable this behaviour). This is easy to implement, but as a temporary solution I like your approach (extending the LCE map) better: if someone simulates a source, at least we'll have the below-the-cathode events in with some not-completely-unrealistic hitpattern. If we ignored the below-the-cathode events completely people might conclude the lone-S1 background is much less significant than it actually is.

Issue #204, opened 2015-08-05
Acquisition monitor (e.g. sum wf between events) 


tunnell:
I need to implement two things:

Written to raw data  - The acquisition monitor is just a board that we will know from the run db.  The MongoDB untriggered reader plugin should fetch this data outside of the event window.  This will also mean adding to the Event class a field called 'time_to_last_trigger' or equivalent.  This is fine since our coincidence pretrigger runs serially.  This means that every XENON1T event will have raw data outside of the event window that needs to be ignored.  

Any analysis - At the moment, we don't know how we're going to use this data.  It's sort of a smart self-veto cut in some cases.  It also records things like the busy.  After we turn on XENON1T, we'll start finding uses for this information that should be known at the analysis stage.  This should remain possible.

JelleAalbers:
We discussed this briefly at the collaboration meeting. For reference: the acquisition monitor is a separate digitizer which will read out several signals such as a GPS clock or the muon veto trigger. Self-triggering will almost certainly have to be used on it to avoid crazy data rates. For more info, see slide 7 of https://xecluster.lngs.infn.it/dokuwiki/lib/exe/fetch.php?media=xenon:xenon1t:org:collaborationmeeting2016_daq.pdf

To start we could just save the AqM self-triggered pulses, maybe just noting the origin in a more sensible way (i.e. instead of saving "this pulse is from channel 6, module bla" but "this pulse is from the muon veto trigger"). For the sum waveforms there may be a lot of pulses though, then we should think about doing a more advanced reduction.

Since the data stream is decoupled from the pax events, it would make sense to store it in the same way as the trigger data. This will go first to MongoDB (so it can be monitored online), then to a to-be-specified file format (currently hdf5, could also be pickle/json/etc, will not be ROOT).

The actual querying can be done by the event builder or by a completely separate application. We can't just treat them as regular channels, we'll have to do a different query: unlike for TPC pulses we do want to read and save the raw data all the time.

PR #392, opened 2016-06-29
Read out acquisittion monitor channels

JelleAalbers:
This configures pax and the event builder to read out the acquisition monitor channels (see #204). The current acquisition monitor map is:
* Module 167, channel 0 = "PMT" 254: analog sum waveform
* Module 167, channel 1 = "PMT" 255: busy signals. The busy spikes once or twice in very rapid succession (few samples) depending if it turns on or off.
* Module 167, channel 2 = "PMT" 256: high-energy veto. 
Each of these acquisition monitor channel gets its own separate 'detector' in the pax configuration. They also have a gain attached so the sum waveforms actually appear in the event display, although this is now just something I chose to make the plots look reasonable. For the analog sum there probably is a correct value -- does anyone know what it is?

After merging this pull request, the pulse rate in each of these acquisition monitor channels should appear in the trigger monitor; the 'other channels' readout will move to "PMT" 257 automatically. The acquisition monitor data will also appear in the event display:

aqm_readout_2

aqm_readout

As you can see, the busy pulses are delayed by several us, and the analog sum waveform may be slightly delayed as well. If we can't remove this delay, we could correct for it in the event builder. The amplitude offset in the analog sum waveform for some channels is because pax attempts to baseline the pulses in the same way as it does regular PMT pulses, but there is not enough pre- and post- trigger to make this make sense.

I imagine the most important use cases for the acquisition monitor are:
- Rejecting events with a busy/he-veto just before it, and for most analysis probably also events with an interior busy.
- Computing the dead time, i.e. time when the busy/he-veto is active, in the trigger.

There are some outstanding issues to solve before we can do this:
* If possible, the busy-on and busy-off channels should arrive in separate channels. Just sending a slightly different pulse means the trigger has to query and parse each pulse, then recognize the waveform shapes. This will probably be complicated, computationally infeasible and prone to bugs. @buetikofer do you think we could separate on- and off- channels?
* We should store a 'time until last busy' and '... high energy veto' in each event. This is a task for the trigger. 

**Warning:** if we merge this, and take data with it, it cannot be processed anymore by the previous configuration of pax (since, to pax running with old config, there will be data in channels it does not know). This means we need to update the pax version used for processing right away after merging. If you want I can make pax more forgiving of unknown channels (like the trigger / event builder) so we don't have this problem in the future.

### DIFF
--- a/pax/config/XENON1T.ini
+++ b/pax/config/XENON1T.ini
@@ -204,8 +204,8 @@ s2_patterns_zoom_factor =             2
 # Since the peaks noted in that figure add up to 2.9%, but the total afterpulse rate is quoted as 4.9%,
 # I (Jelle) arbitrarily added a 2% probability population with 1us decay time.
 # TODO: Calibrate this more properly once we have data
-# TODO: This should soon become part of the pmts dictionary (/database?), then we'll be able to add PMT-dependent info.
-# The original photon is not yet deleted when an afterpulse is made (but this is such a small effect we probably don't care
+# TODO: Add PMT-dependent info?
+# TODO: The original photon is not yet deleted when an afterpulse is made (but this is such a small effect we probably don't care about)
 pmt_afterpulse_types = {
     'H2':  {'p': 0.002 * 1.8/4.9,
             'time_distribution': 'normal',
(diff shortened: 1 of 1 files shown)


JelleAalbers:
Thanks for the info Lukas! The 31.25 factor is close to what I put before just to make the waveform roughly match; I've put in the your value now.

When you have separate busy / HE-Veto on and off channels let me know, then I can implement a dead time calculation in the trigger.

I've talked with Chris and will merge this after it builds so we start taking this data. There is still some work to do in pax, (e.g. storing time until previous & next busy for each event), so I won't close #204 yet.

sanderbreur:
Lukas, I would say that we do not want to wait with implementing the
separation of the on- and off- channels as we need that to calculate the
total dead time.

So I would go with the short guide.

S

Op wo 6 jul. 2016 om 17:53 schreef Jelle Aalbers <notifications@github.com>:

PR #396, opened 2016-07-18
Fixes for muon veto compatibility

JelleAalbers:
These are a few small fixes to ensure the muon veto is again supported by the event builder. I was able to build events with the command:

    event-builder --secret_mode --run_name 160718_0928 --detector muon_veto

You can try leaving off the --secret_mode, which might start the data flow outside LNGS (depending on cax of course). You also probably want different trigger settings; the easiest way is to use the --force_multiplicity_trigger option, but for more flexibility, see https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:aalbers:trigger_upgrade#the_trigger_decision.

I've disabled looking in the runs database for processing/trigger configuration overrides for the muon veto for now. That means it always uses whatever is in XENON1T_MV.ini unless you pass a command-line override.


tunnell:
I have no problems merging when you want.  Just @CyrilGrignon tell me when you get it working.

PR #401, opened 2016-07-28
TopPatternFit ring fix

pelssers:
This PR fixes the ring structure seen in the TopPatternFit reconstruction.

**In short:**
This solution modifies the S2 per-PMT LCE maps for PMTs 35-65 (the second ring). These maps show relatively high LCE values just beyond the TPC radius causing events below this PMT ring to be reconstructed too far outward. By trimming 4mm of these LCE maps this problem is solved. The reconstruction is more uniform, has better resolution and still all the desired properties (such as coverage of the confidence regions). This solution does however prevent the algorithm from reconstructing in the outer 4mm of the TPC.

**The radial distribution of events, old LCE maps:**
radial_dist_original_log
**The radial distribution of events, new LCE maps:**
radial_dist_trim2_log

**More details in this note:**
https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:pelssers:posrec_compare_1t#fixing_toppatternfit_bug_ring_feature


tunnell:
Thanks.  How realistic is it just to fix the LCE maps in Geant4?

pelssers:
The LCE maps from the Geant4 MC are correct, the problem is that the region where the S2 signal is generated is slightly larger in radius (2 mm) than the main TPC volume.
This means any photons simulated in this 2mm will almost all reflect upwards to the top PMTs leading to the higher LCE.
So the map is physically correct but the higher LCE still throws off the position reconstruction since a real signal (electrons drifting up from the main TPC volume) will never end up in this region and should not be reconstructed there.

weiyuehuan:
Hi Bart,
Thanks! I am stilled worried about this trim correction. After the 
electrons being extracted from liquid to gas xenon, it will end up on 
the wires (mostly cross points) on Anode electrode. so the S2 region 
should be a bit larger than main TPC volume. I attached a plot I made 
years ago, it shown the electrons trail (color means the photons 
produced). Anyway we will cut away this events on the edge, but I wonder 
where it will be reconstructed if the real signal happened in the outer 
2mm ?

Best regards,
-Yuehuan

pelssers:
Hi Yuehuan,
So if I understand correctly, electrons can end up in the larger region beyond the main TPC radius. In that case I agree that we should be able to reconstruct them there.
At the moment (with this modification) any event beyond r = 47.5 (which I called the LCE radius) is reconstructed just within this radius.

This can be seen in the below plot, the increase in the number of events close to the LCE radius is due to events from further out that are reconstructed here.
radial_dist_trim2_log

I'm sorry I do not see the attached plot you mentioned in your mail, could you try again please.

Bart

weiyuehuan:
Hi Bart,

Thanks, Yes, this is my meaning. I attached plot again.
I understood now. All the events beyond r 47.5mm will be reconstructed 
within this radius. So it means most edge events will be reconstructed a 
bit inward because main TPC volume radius is ~47.9 and S2 region is a 
bit larger. Don't know how much of this effect.
In Figure U6b in you note, maybe you can set X-axis as (R_recon - 
R_real) and SetLogy that just make sure we can't reconstruct the edge 
events to fiducial mass.

Best regards,
-Yuehuan

JelleAalbers:
Hi Bart, Thanks a lot for fixing this problem. I think we should merge this and reprocess at least the recently taken data quickly, so we can see if there is any drift field inhomogeneity.

I'm not sure I understand the details yet though (you've tried to explain it to me before, I probably forgot..).

Why are the GEANT patterns different from the LCE map-generated / fax patterns? You say the LCE maps are a correct representation of the light emission from these outer regions, so then the patterns should be the same, whether or not S2 electrons actually end up there (or how often they end up there), right?

tunnell:
We need to discuss Monday before merging since this affect many things

tunnell:
Pull with master?  Then I think we can merge after builds.

PR #406, opened 2016-08-07
Trigger TPC when muon veto triggers

JelleAalbers:
This adds the muon veto trigger channel on the acquisition monitor as channel 259 in pax. It also lets the TPC trigger whenever the muon veto triggers (currently a few times per minute). 

The events are distinguishable in the analysis by having a trigger signal with a typecode of 106 (the muon mass in MeV). 

I tested this for a few minutes on Sunday when the fields were off. It works, even though I did not catch an actual muon during those few minutes:

mv_trigger

I also refactored a few things in the trigger:
  * The trigger's name for pulses used to be 'times'. I changed this in most places.
  * I pulled the conversion from pmt number & ADC module to PMT number into a separate trigger plugin. This makes trigger.py itself simpler.
  * Sorting the data is now done inside the trigger, not in the MongoDB reader. It used to be done in parallel in each of the query threads, but since we adopted split hosts that is no longer possible (except if we make some complicated chained thread system). Perhaps in this new place we can profile it more easily; it certainly takes less lines. 

When we merge this, we should also take care to:
* Release a new pax, and up the version used for processing. As with other changes that affect the raw data, you will not be able to process data taken after this change with previous versions of pax. 
* Insert an extra 1e5 in the gains override list in the run doc for at least the runs taken with this mode (it's ok to do it for all the runs; extra gains won't cause problems, missing gains will.)
* Squash the commits... sorry about that.


pax/trigger_plugins/DecideTriggers.py
```diff
@@ -42,3 +44,29 @@ def flag_triggers(signals, p_matrix):
...
+    """Make a triggering signal for EVERY individual pulse in a particular detector"""
+
+    def process(self, data):
+        for detector, typecode in self.config['typecodes_for_detector'].items():
+            channels = self.trigger.pax_config['DEFAULT']['channels_in_detector'][detector]
+            pulses = data.pulses[np.in1d(data.pulses['pmt'], channels)]
+            if not len(pulses):
+                continue
+            self.log.debug("Triggering on every individual %d pulses from detector %s" % (len(pulses), detector))
+
```
tunnell: __[flake8]__

PR #415, opened 2016-08-29
Change PMT alive threshold.

pelssers:
The TopPatternFit position reconstruction plugin uses the gain of the PMTs to determine whether they are on or off. PMTs that are off should have their gain set to 0. PMTs that are on typically have gains of the order of 10^6.
It was found (by Jelle and me) that some "off" PMTs have non-zero gains with typical values of around 10^-22.
To also correctly label these PMTs as off this pull request changes the threshold from 0 to 1 allowing TopPatternFit to reconstruct peaks with the correct list of contributing PMTs.


ErikHogenbirk:
Just for reference, I think is also the cause of some peaks with area ~10^28 (!) in data. In case that a large peak gives _some_ signal in one of the 10^-22 gain PMTs, this is instantly in this range. I suggest merge soon to avoid confusing people who stumble upon this in some way.

tunnell:
Thanks

Issue #419, opened 2016-09-07
Mode for not triggering


JelleAalbers:
It would be nice if there was a mode that made the trigger not trigger on anything, but still clear the data and do the monitoring as always. You can get this behaviour with the current settings, e.g. `"trigger_probability": { '0': {'9999':1}, '1': {'9999': 1}, '2': {'9999': 1} }` in the Trigger.DecideTriggers section together with `signal_separation = 1` in the Trigger section, but this messes up the trigger signal monitoring. It would be nicer if there was a simple flag for this.

JelleAalbers:
You now have `do_not_trigger` option in Trigger.DecideTriggers. I made a background_stable_donttrigger runmode that uses this.

Issue #418, opened 2016-09-07
Do not hardcode host locations in trigger, parse run doc field instead


JelleAalbers:
https://github.com/XENON1T/pax/blob/master/pax/plugins/io/MongoDB.py#L79

Issue #412, opened 2016-08-12
Trigger should have a heartbeat


JelleAalbers:
It should periodically insert or update something in MongoDB to indicate it is still running. That would allow the website to display some status if the trigger is actually running.

coderdj:
How hard would it be to have it put a doc with timestamp, current run, and triggers/second?

Issue #407, opened 2016-08-08
Event builder should set data entry to 'error' on crash


JelleAalbers:
Right now it only sets run_doc['trigger']['status'] to error when it crashes. Since the website doesn't check for this, the runs never gets an error mark, and the error may go unnoticed.

PR #420, opened 2016-09-08
Several trigger issues

JelleAalbers:
This addresses several trigger issues:

* #418: read hosts from the run doc (reader.ini.mongo.hosts)
* #419 : an option `do_not_trigger` in the Trigger.DecideTriggers section that ensures no events are created at all.
* #412: The trigger periodically (every time it processes a 21.47sec batch of data) inserts a document with `data_type='batch_info'` into the trigger monitor collection for the run it is working on. This includes the timestamp and various useful info about that batch (such as number of events built, total duration of them, etc). This is not completely what @coderdj asked for: the website still needs to figure out what run the trigger is working on  -- but querying for runs with a trigger status of 'staging' or a raw data which is 'transferring' on xe1t-datamanger should accomplish this. Some old runs where the trigger crashed may also have this, so you probably want to take the latest one, or limit it to a few days old.
* #407 when the trigger crashes in the middle of a run, it will now also set the data entry in the run doc to error (as opposed to leaving it to 'transferring'). 

None of these have been tested yet.


tunnell:
once tested, can merge.

tunnell:
I mean tested on the DAQ.  Not Travis.

JelleAalbers:
I think I found the offending bug in the trigger dead time calculation (#416), hope this will work...

JelleAalbers:
This will be rolled out on the DAQ today or tomorrow.

PR #421, opened 2016-09-13
Used the correct decoder for the muon veto data

CyrilGrignon:


JelleAalbers:
Small MV configuration change, no impact on rest of pax.

Issue #422, opened 2016-09-14
Save total number of pulses (per channel) in datastructure


JelleAalbers:
May be useful for debugging or to get an idea of what the trigger is doing. It's very cheap in any case, just one event-level field.

PR #425, opened 2016-09-15
Pulses per channel

Riciardos:
Pull request for issue #422.

Don't know what the exact purpose is, but it's in the datastructure now. The pulses are counted in the CheckPulses before any truncating is done, this way it represents better what the trigger is doing I guess.


JelleAalbers:
Pax 5.7.0 will be released today or tomorrow.

Issue #423, opened 2016-09-15
Better event display peak label hiding


JelleAalbers:
Currently the event display shows the labels of all peaks in the waveform. However, many of our events have a large single electron tail; if each of these are labeled, an ominous black cloud forms:

![000009](https://cloud.githubusercontent.com/assets/4354311/18554789/a5012502-7b65-11e6-9ec9-3b06ea74d716.png)

It would be nice if we labeled only the largest few peaks, or do something smarter to label a maximum number of peaks per few us. There is already an option `hide_peak_info` you can add in the `Plotting` section to not show any peak labels.

JelleAalbers:
Addressed in #427

JelleAalbers:
Sander and I tried the fix in #427 a few weeks ago, but nothing happened. Maybe we did something wrong though, we or someone should probably check again.

JelleAalbers:
We don't know if this is fixed yet

Riciardos:
I have taken a look at it and for me it seems to be working for me on the XENON100 file in the data folder. You just have to include in the XENON100.ini file in the [Plotting] section a num_of_sections variable and an annotation_filter list like:
`[Plotting]`
`num_of_sections = 40`
`annotation_filter = ['largest']`
(it should work the same in the XENON1T config file)
Instead of `['largest']`, you can also try `['largest_s1s', 'largest_s2s']`, so it will filter the types separately (so you won't miss an s1 sitting right next to an s2).

Also, I can understand if having to edit the config file every time could be a bit cumbersome, so I can also try to make it so you can override it with a command line option like: `--annotation_filter ['largest', 5]` to include the largest 5 S1s and largest 5 S2s.

I will create a separate working branch so you can check out if it's still not working for you.

Riciardos:
There is a branch 'annotation_filter_working' with how it's working for me, so check that one out and see if it's working for you as well (I also cleaned up a print statement and a comment). Only thing I've added was the ['largest'] list to the annotation_filter variable in _base.ini

Riciardos:
There is a branch 'annotation_filter_working' with how it's working for me, so check that one out and see if it's working for you as well (I also cleaned up a print statement and a comment). Only thing I've added was the ['largest'] list to the annotation_filter variable in _base.ini

PR #426, opened 2016-09-18
fax update

mcfatelin:
The main modifications are

1) Adding one configuration for simulation: gauss_noise_sigmas. This is a list of baseline fluctuations on each PMT channel. If "gauss_noise_sigmas" is None, fax will use "gauss_noise_sigma" as an average baseline fluctuation over all PMT channels for waveform simulation.

2) Adding the g4_id ( eventid in input G4 root file) into fax_truth file. This is what we discussed in Tuesday's telecon for helping MC-data comparison. If the input is csv file, the g4_id is -1.

PR #427, opened 2016-09-21
Fixed the black blobs of annotations in event plotter. 

Riciardos:
Fix for issue #423.

The timeline of the event is split into sections, and the peaks are placed in the approriate time section.
If a section contains more than or equal to 1 peak, add the biggest peak to the annotation list.
The ones in the annotation list will be annotated, the rest are skipped.
Configurable from config file: set less_black to True if you want this fix.
Number of sections also changable from config: lower number should lead to less black blobs(as more peaks are skipped per section)


pax/plugins/io/Plotting.py
```diff
@@ -179,6 +206,11 @@ def plot_waveform(self, event,
                                       connectionstyle="angle3,"
                                                       "angleA=0,"
                                                       "angleB=-90")
+                if less_black:
+                    #Go to next peak if not the biggest in section, otherwise annotate!
```
tunnell: __[flake8]__
* [E265] block comment should start with '# '

JelleAalbers:
Hi Ricardo,

Thanks again for looking into a pax issue! This is an important improvement for the event display.

Sometimes it is quite useful to annotate peaks even when they are close together: e.g. when an S1 and an S2 are close together, or when we have two large S2s near to each other. Maybe you could complement your fix with an option (on by default) to always annotate the largest X peaks, or the largest X peaks of each peak type, whether or not they are the first in their 

Make you can show an example event display after your fix? You can drag & drop images into the text editing field when you're writing a comment.

(if you fix the lint errors, the Travis builds might still fail; some dependency conflict apparently emerged recently. I hope it gets resolved upstream, else we'll have to eg. pin/downgrade matplotlib as Chris suggested here: https://github.com/XENON1T/pax/pull/426#issuecomment-248409738)

Riciardos:
Hey Jelle,

I have changed some stuff around and I think you will like it this way. I re-factored the annotation_filter into a separate function which reads out a list of options from the config file. 

The main S1 and S2 are always annotated, if they are present.
`'largest' ` will do what it did before: only annotate the largest peak in a section
`'largest_s1s'` or `'largest_s2s'` : will do the same but only for s1s or s2s respectively
`'all_s1s'` or `'all_s2s'` : will annotate all s1s or all s2s (duh)

and any of these can be combined, so for example if you want all the s1s annotated but only the largest s2s, you can set `annotation_filter = ['all_s1s','largest_s2s']` and it will do exactly that. 
Just setting it `'largest'` will give the best readability, and like I said earlier, the main S1 and S2 are always annotated.

I don't have access to Xenon1T data so I had to use the example file in pax, and here is the first event without any filter:
even1nofilter

and here it is with the 'largest' option enabled

event1withfilter

Also for the Travis stuff, I did change the .yml file, but I think it might have something to do with the fact I imported `attrgetter` from the operator module, which is not a built-in function,but the module should be included in every Python environment. Anyway, it's checking it now and we'll see what happens.

tunnell:
@JelleAalbers you find with this?

JelleAalbers:
It would be good to add an option to never hide the largest (say) 5 s1s and s2s in the event, otherwise we may not spot a double scatter.

But I guess I won't have time to look at that for the moment, and this definitely improves the event display. Right now the peak annotations are almost useless in high-energy events. So we can merge this and add the option to not hide the largest x S1s / S2s another time.

tunnell:
Approved. Good, let's merge.  Can always improve method later.

Warning: if two peaks too close, hides label.

sanderbreur:
Ricardo thanks for making this possible.

S

Op di 25 okt. 2016 om 12:53 schreef Christopher Tunnell <
notifications@github.com>:

PR #428, opened 2016-09-23
Waveform simulator S2 luminescence model update

JelleAalbers:
This introduces a new S2 luminescence model which:
  * Takes into account the speed-up of electrons as they get near the anode
  * Allows for an (x, y) dependent gas gap, to be able to reproduce the XENON1T single electron width results

The settings for XENON1T waveform simulation are also updated to better match the data we saw at cathode = - 15 kV. I will present notes on this matching later.

Note this new model probably breaks the XENON100 single electron matching -- but that was done to run 10 data anyway, and likely doesn't hold for later data. 

More information will appear in notes over the coming weeks.


tunnell:
Conflict

tunnell:
Assume you want to present notes first (this was first one today)?

JelleAalbers:
Yes, I'll send around another note on the single electrons shape, probably next week.

JelleAalbers:
I've included the latest values based on my two notes:

* https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:aalbers:drift_and_diffusion
* https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:aalbers:single_electron_waveform_shape

With one exception: I've left the transition time spread at 8.8 ns rather than the fit value of 42 ns. The fit is probably influenced by distoring factors (such as mixing of populations from different regions due to the poor posrec resolution at these low statistics), and this setting mainly impacts the S1 shape (which hasn't been studied yet).

For the warping of the anode, the data-extracted map is used that I also show in the note:

warping_map

Where for definiteness I've defined 0 to be the gas gap according to 5mm - the average level meter readout.

Change to z-computation 
----------------------------------
I included one change to regular processing: the z-computation now takes into account the drift time offset of 1.7 us due to the liquid above the gate. So z is now computed as:

    ia.z = - self.config['drift_velocity_liquid'] * (ia.drift_time - self.config['drift_time_gate'])

Note this will lead to positive zs for events above the gate.

PR #429, opened 2016-09-27
Reformat trigger data to use list

coderdj:
It would make the trigger monitor frontend much easier if we implement this change. The idea is to save trigger data as arrays, which can be queried and manipulated in the aggregation pipeline and not as compressed ndarrays, which generally have to be fully unpacked before they are useful.

A side effect is that the format will change for trigger monitor data stored offline. We will have to put a small update to hax to test what the format is so it can read properly.


JelleAalbers:
This will be rolled out on the DAQ today or tomorrow.

PR #431, opened 2016-10-06
New per-PMT S2 LCE maps

pelssers:
I converted the new per-PMT S2 LCE maps made by Jialing to the pax format (using XeAnalysisScripts/mapmaking) and set the s2_mean_area_fraction_top to the new value of these new maps (old value 0.68, new value 0.55).

Information about the maps can be found here (note by Jialing):
https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=jialing:s2mapwithupdatedgeo

The new LCE maps improve the position reconstruction as described here:
https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:pelssers:posrec_compare_1t&#fixing_toppatternfit_bug_ring_feature_using_new_s2_lce_maps


tunnell:
Great to see this work!  Can we discuss it on Monday?  @dlingling and @pelssers ?  I don't understand it to be honest.  If the LCE maps aren't that different, then how could the ring disappear?  It also looks like there is an underdensity right inside of the TPC radius.

pelssers:
Hi Chris, yes, let's discuss this on Monday.

dlingling:
Sure. It's an interesting topic.

Best,
Jialing

PR #436, opened 2016-10-17
S1 lce kr83m sept29

berget2:
Updated the S1 LCE map from the Sept. 29 2016 Kr83m update. The primary difference is that an updated processing of the data provided the correct drift velocity, and thus z positions, that are used to build the LCE map.

original note: https://xecluster.lngs.infn.it/dokuwiki/lib/exe/fetch.php?media=xenon:xenon1t:berget2:xe1t_kr83m_eventselect.html

conversion to pax format: https://github.com/XENON1T/XeAnalysisScripts/blob/master/Kr83m_1T/figures/Xe1T_LCE_MapConvert.ipynb


JelleAalbers:
Thanks for converting the maps Ted! We'll include this in the next pax release, probably this week.

tunnell:
Approved.

PR #437, opened 2016-10-17
Compute saturation correction even if not included in total

JelleAalbers:
This ensures that the S1 and S2 correction values computed by Sander's saturation correction algorithm (see https://github.com/XENON1T/pax/pull/237) are saved even if we do not include them in the total correction (interaction.s2_area_correction and interaction.s1_area_correction).

We do not currently include the saturation correction for XENON1T since they're a bit more experimental than the other corrections (although they work fine for XENON100, they rely on the Monte Carlo LCE per-PMT map). Hopefully after we see they work well on a couple of datasets we can include them in the total correction, as we do for XENON100. You will always be able to access all individual corrections separately (see https://github.com/XENON1T/pax/pull/322).

### DIFF
--- a/pax/config/XENON100.ini
+++ b/pax/config/XENON100.ini
@@ -166,18 +166,19 @@ s2_pairing_threshold = 70  # pe
 
 
 [BuildInteractions.BasicInteractionProperties]
+include_saturation_correction = True
+
 # Add 'zombie PMTS' here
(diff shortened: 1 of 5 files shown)


tunnell:
OK, approved.  Let's get per PMT LCE maps from Kr83m instead of MC before we use implement this into S1 by default.

PR #440, opened 2016-10-25
Show all PMTs in hitpattern plots by default

JelleAalbers:
This changes the PMT hitpattern plots in the pax event displays (--plot and --plot_interactive) to show also the PMTs that were not hit during the event:

show_all_pmts

To return to the previous layout (showing only pmts that are hit), you can include the option show_all_pmts=False.


tunnell:
Depends #442.  After that builds and merges, should be fine.

tunnell:
Approved.

PR #443, opened 2016-10-25
Generalizing naming for MC input to fax

pdeperio:
* Rename "NEST" to "MC" to accomodate multipleS1/S2 generation methods (e.g. Geant4-NEST, nSort)
* Revert branch names "Nest_" to "Fax_" to unify format (as above)


tunnell:
Updated your branch, no objections from me.  I'll approve so you can merge once build passes.  In the meantime, Jelle and others can object.

PR #447, opened 2016-11-02
Fix data model using Decimal in wrong case

tunnell:
Don't use decimal in lists if decimal false

### DIFF
--- a/pax/data_model.py
+++ b/pax/data_model.py
@@ -159,8 +159,10 @@ class Model(object):
                                         use_decimal=use_decimal) for el in v]
             elif isinstance(v, np.ndarray) and convert_numpy_arrays_to is not None:
                 if convert_numpy_arrays_to == 'list':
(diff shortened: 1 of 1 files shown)


pax/data_model.py
```diff
@@ -159,8 +159,10 @@ def to_dict(self, convert_numpy_arrays_to=None, fields_to_ignore=None, nan_to_no
                                         use_decimal=use_decimal) for el in v]
             elif isinstance(v, np.ndarray) and convert_numpy_arrays_to is not None:
                 if convert_numpy_arrays_to == 'list':
-                    result[k] = [decimal.Decimal("%f" % a) if isinstance(a, float) else a for a in v.tolist()]
-                    result[k]
+                    if use_decimal:
+                        result[k] = [decimal.Decimal("%f" % a) if isinstance(a, float) else a for a in v.tolist()]
+                    else:
+                        result[k] = [a for a in v.tolist()]
```
JelleAalbers: You don't need this list comprehension, can just be v.tolist()

tunnell:
simplified

PR #450, opened 2016-11-07
Calibrated x1t clustering

JelleAalbers:
This changes pax's clustering settings to be more appropriate for the current conditions in XENON1T. A description and discussion is here: https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:aalbers:clustering_calibration

It includes a few smaller changes:
  * Change the TPC length to the more accurate value computed by Michelle
  * Include the old clustering settings temporarily via XENON1T_oldclustering.ini
  * Include a XerawdpImitation_1T.ini which allows you to run the XerawdpImitation on XENON1T (with rather awful results as Xerawdp isn't calibrated to 1T). This is more a fun feature than a useful one.
  *  Small fix that enables `--plot_interactive` to output to a directory if you specify `--output mynicedirnameforplots`


tunnell:
Review good, we can merge before release.

PR #453, opened 2016-11-08
First attempt at writing out fax truth to ROOT file

pdeperio:
(Sorry, more ROOT garbage =P.) Trying to better unify the fax truth info to the current output files from pax/hax, i.e. to have a fax truth file with exact event-by-event correspondence to the pax/hax ROOT trees.

This pull request is the simplest and fastest I could do without more complicated (but nicer) stuff as suggested by Jelle below. Therefore, it is still just a tree with all peaks across all events in each row, as in the attached file:
[faxtruth.root.zip](https://github.com/XENON1T/pax/files/578627/faxtruth.root.zip)
and will require an external adaptor to get it into the final form. This adaptor shall be easily added to any production chain.

This is to give analysers additional flexibility, so the existing .csv will remain too.

This requires a ```conda install root_pandas```

And some suggestion from Jelle, but probably not on the timescale of the upcoming mass production:

> I think there's a few approaches:
 
> (a) If you want something like the Xerawdp root file where you have a vector of peak areas, widths etc, you'll probably have to write your own code with pyROOT rather than use root_numpy or root_pandas. Numpy and pandas don't handle ragged or nested datastructures very well. But you probably don’t want this anyway because it’s hard to compare to the regular pax format.
 
> (b) If you want to have the same format as a pax root file, the easiest way is to create pax events and fill them with peak objects (and maybe later on some more advanced truth info like hits and interactions). Then, you can have a new Processor instance running inside your plugin, which runs EncodeROOTClass and WriteROOTClass (and no input plugin); you can feed the events with truth info into that with Processor.process_event. It would involve a bit of pax magic but it should work.
 
> (c) You could also just keep the output as it is, peak-level only, and compare against the hax PeakExtractor output or against pax's TableWriter output (what you get with --output_type csv or --output_type hdf5). This is what I do in PeakfinderTest (peakfindertest is actually older than the pax root output).
 
> I like (c) as it keeps the output as simple as possible. You can still group the peaks by event later (with pandas' groupby or other tools) and you don't have to deal with complicated datastructures.
 
> If you want to store more complicated info I think (b) is your best bet, since then you end up with the same format as pax. If we ever manage to switch the root format out for something more stable (though that's probably for the "next generation"), you can also switch over to that easily by just changing which output plugins your mini-processor runs.


pax/plugins/io/WaveformSimulator.py
```diff
@@ -12,6 +12,7 @@
 
 from pax import plugin, units, utils
 
+import root_pandas
```
tunnell: __[flake8]__
* [F401] 'root_pandas' imported but unused

pdeperio: Actually, this is needed to endow DataFrame with the to_root function:
https://github.com/XENON1T/pax/pull/453/commits/5561d5fffe7f5d169cb2324768d538b3a9a97e42#diff-9b725188cd6edbaadd4665a5c76f69ffR57

JelleAalbers: You can add a `   # noqa` comment on the same line to make the code checker ignore it

pax/plugins/io/WaveformSimulator.py
```diff
@@ -12,6 +12,7 @@
 
 from pax import plugin, units, utils
 
+import root_pandas # noqa
```
tunnell: __[flake8]__
* [E261] at least two spaces before inline comment

pdeperio:
@tunnell @JelleAalbers how to install ```root_pandas``` for the travis tests?

tunnell:
Edit `.travis.yml`

pax/plugins/io/WaveformSimulator.py
```diff
@@ -12,9 +12,11 @@
 
 from pax import plugin, units, utils
 
+import root_pandas  # noqa
 
 def uniform_circle_rv(radius, n_samples=None):
     """Sample n_samples from x,y uniformly in a circle with radius"""
+
```
tunnell: __[flake8]__

Issue #375, opened 2016-06-14
Improve baseline computation robustness


JelleAalbers:
Truncated pulses near the start and end of the event sometimes have a hit inside the regions used for baseline computation, skewing the baseline upwards.

We currently use a rather non-robust way to compute the baseline anyway (the mean), it would be better to switch to the median instead. Unfortunately that means we can no longer do it in numba (unless somebody wants to code up a median computation algorithm), but I doubt the speed reduction would be such a big issue.

JelleAalbers:
Addressed by #479 (includes changes that were in #455)

PR #455, opened 2016-11-10
Separate baseline computation and hitfinder, MessagePack

JelleAalbers:
This includes two things:
  * Move the pulse property computation from the hitfinder to a separate `PulseProperties` plugin. Since we do this before truncating pulses outside the event, this solves #375.
  * Introduce MessagePack data output support. This is slower to write than pickle but faster than BSON, and more compact than either. We should move away from pickle in any case, see https://www.youtube.com/watch?v=7KnfGDajDQw.

However, the main motivation for this is shrinking the raw data size by a factor 2, as explained in the note here: https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:aalbers:raw_data_shrinking&#examples_of_pulses_after_baseline_data_removal 

The `shrink_raw_data.ini` activates the shrinking procedure. If we decide we want to do this in the event builder I can change the eventbuilder config as well. If we go down that route it would take about 50% longer to build events.


JelleAalbers:
This PR will rise from the ashes shortly, when I make a pull request for the new hitfinder integration behaviour building on this.

PR #457, opened 2016-11-11
Neural network update

pelssers:
This PR adds two neural networks to PAX. These networks have been trained by Yuehuan, one network with three PMTs off in the top array and one network with seven PMTs off in the top array.

If merged this will enable the neural net with 7 PMTs off.

Details about the training and performance can be found in Yuehuans note [here](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:sim:notes:yuehuan:position_reconstruction_geoupgrade)

I've also updated my conversion script (from Yuehuans c++ weights and biases file to a PAX friendly .npz file) [Neural Net Tools](https://github.com/XENON1T/XeAnalysisScripts/tree/master/NeuralNetTools) which contains a list of all available neural networks and their settings and options.


tunnell:
Great, sounds good.  One thing: the PMT conditions will change throughout the lifetime of the experiment so we don't want one NN for all runs at all times.  It would be useful if you could overload through the rundb (find_and_modify) the filename where the NN is found.  That way we can have things like less than run 1000, use this NN, after that, this other one.  Just by changing file name.

PR #461, opened 2016-11-18
Use MongoProxy to handle autoreconnect exceptions

JelleAalbers:
This makes the rundb code in pax robust to NotMaster exceptions; if they occur, we retry a few times; if the operation fails for more than two minutes, the exception is reraised.

Currently we're experiencing regular "NotMaster" exceptions, triggered by mongodb when the runs db replica set enters election season. These are "Autoreconnect" exceptions, meaning you can retry the operation later. Pymongo will NOT do this (so autoreconnect is a [misleading name](https://jira.mongodb.org/browse/PYTHON-197)), the [user code has to do it](http://api.mongodb.com/python/current/examples/high_availability.html#handling-failover).

We don't want to wrap every call to the runs db in pax in a try-except loop. Fortunately, someone has written a proxy class which makes mongodb do this automatically; see https://github.com/arngarden/MongoDBProxy and http://www.arngarden.com/2013/04/29/handling-mongodb-autoreconnect-exceptions-in-python-using-a-proxy/. 

https://github.com/arngarden/MongoDBProxy/blob/master/mongo_proxy/mongodb_proxy.py#L69

I have no idea if this works, we would have to test it first.


JelleAalbers:
@tunnell any objections? This has been running on the event builders for some time. 

I removed mongoproxy from requirements.txt. If you do not install it or it fails to import, we will simply fall back to vanilla mongo.

tunnell:
No objection


From: Jelle Aalbers <notifications@github.com> <notifications@github.com>
Reply: XENON1T/pax
<reply@reply.github.com>
<reply@reply.github.com>
Date: 7 December 2016 at 09:43:22
To: XENON1T/pax <pax@noreply.github.com> <pax@noreply.github.com>
Cc: Christopher Tunnell <tunnell@uchicago.edu> <tunnell@uchicago.edu>,
Mention <mention@noreply.github.com> <mention@noreply.github.com>
Subject:  Re: [XENON1T/pax] Use MongoProxy to handle autoreconnect
exceptions (#461)

@tunnell <https://github.com/tunnell> any objections? This has been running
on the event builders for some time.

I removed mongoproxy from requirements.txt. If you do not install it or it
fails to import, we will simply fall back to vanilla mongo.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/XENON1T/pax/pull/461#issuecomment-265481598>, or mute
the thread

PR #471, opened 2016-12-20
Realistic noise and ZLE thresholds

JelleAalbers:
This includes a small (5MB) file with XENON1T noise data to provide example noise to simulate upon. 

Whether this is sufficient depends on your purpose:
  * If you want to check the distortion of areas or the effect on the hitfinder acceptance, any somewhat realistic noise will do; you care less about e.g. rare small spikes in the noise. 
  * If you want to check for generation of fake S1s from noise, it is definitely insufficient, and you want to look at a much longer time range and/or make a harmonic model of the noise so you can simulate even larger time ranges.

The noise is taken from a single 1ms noise event (from the most recent noise run, 5037), but it is shuffled for each channel independently in blocks of 480 samples (approximately the periodicity in https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:first_attemp_use_wf_simulator_check_g2_energy_dependence) to provide more variation.

Perhaps more importantly, this updates the thresholds used in the zero-length encoding emulation. Currently this is set to 50 ADC counts, whereas we are already operating with a 15 ADC count threshold (for most channels)... a substantial difference.


pdeperio:
Added the script I think @JelleAalbers used and sent me a long time ago:

https://github.com/XENON1T/XeAnalysisScripts/blob/master/meetings/first_look_xenon1t_noise/Extract%20noise.ipynb

JelleAalbers:
Yes, that's the script. Thanks for adding.

PR #475, opened 2017-01-11
Fax update jan 2017

mcfatelin:
PMT after pulses updated using Julien's analysis, which can be found on Midway under:

/project/lgrandi/julienw/AP_Run_5441.root


mcfatelin:
@JelleAalbers thx for checking. Sorry I just noticed the unit of scale is ns which shall be us in this case. Will update soon.

pdeperio:
Should fix https://github.com/XENON1T/mc/issues/18

PR #478, opened 2017-01-16
Realistic photoelectron pulse model

JelleAalbers:
This adds, and enables, an option to the waveform simulator to use a custom photo-electron pulse model (in place of the default double-exponential one). The model can be specified by a list of times and amplitudes.

My main focus was on making the specification of a model possible, not on getting a super accurate model yet -- that would be a separate study (maybe I or someone else can do this soon). Just to get something reasonable...

waveform_model
[sorry, time on x axis should be us, not ns!]

... I took the median normalized waveform of 729 hits in xenon-light data between 133-166 ADC counts (across all PMTs). I checked the waveform is hardly sensitive to changing these criteria. There are more hits around 70 ADC counts, but as their amplitude is lower, the resolution in the tail is worse. 

The digitizer simulation algorithm is not changed. Although we specified the model as a linear interpolation between 10-ns spaced points, we don't need to do so, and even tough we did, the pulse shape still varies due to photons not always falling on the same place in the digitizer time grid.

Minor changes:
  * The scipy requirement is bumped to 0.17 because we use some features in scipy.interpolate that require them. 
  * The travis pins to scipy/matplotlib versions are dropped; these were needed long ago for some problems, and are now just preventing us from getting bugfixes.
  * Fix for the DeleteLowLevel mode 'all'; it did not actually clear the all_hits event attribute.


pdeperio:
Does using the xenon-light data effectively include effects such as afterpulsing and photon propagation, such that this should not be enabled at the same time as those models? 

(p.s. Looks like you didn't finish your last paragraph about the digitizer.)

JelleAalbers:
Thanks, I've updated the paragraph.

Yes, using any data from the detector we would get afterpulses (or just accidental coincidence of hits). to minimize this influence, I only took pulses with a single hit, and used the median waveform (instead of the mean) to reduce the influence of any small outlier population on each sample. I don't think photon propagation affects us here; I just aligned the hits on the maximum. This will give a small alignment error, which perhaps broadens the narrow main peak a bit. 

It might be a good idea to confirm that the area/amplitude ratio for these hits is the same as observed in data (see e.g. https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:aalbers:first_full_drift_events#photoelectron_amplitude_area_ratio), so we get a realistic hitfinder efficiency in the simulation.

As I mentioned I don't think the model here is great, but I hope it will help us test hitfinder integration biases (and evaluate if things like #479 fix/lessen the observed nonlinearities).

mcfatelin:
@JelleAalbers Good point to have customized photoelectron model available in simulator. However my only comment is which model we use a default in XENON1T.ini. Maybe it is good to put the customized model in a separate simulation configuration instead of the default XENON1T.

JelleAalbers:
I suspect we're going to release a new pax version soon. @mcfatelin @pdeperio Would you like this model to be the new default or stick with the old model for now (which we suspect is quite unrealistic, due to missing the late-time component)?

pdeperio:
Perhaps I'm still not fully understanding what you're doing still. Doesn't part of the "missing late-time component" (tail in your figure above) partially come from photon propagation in the detector? Could you please expand your description of how you prepared that plot?

pax/config/XENON1T.ini
```diff
@@ -188,12 +188,25 @@ pressure =                            1.94 * bar         # Daily check email 16
 temperature =                         (-95.7 + 273.15) * K    # TE105 on 16 October, described as "top of drift region" here: xenon:xenon1t:dsg:cryogenics:tpc:temperature_sensors
 anode_voltage =                       4.0 * kV           # xenon:xenon1t:aalbers:drift_and_diffusion
 
+# Pe pulse model
+    # Old model:
+    #pe_pulse_model            =           'exponential'
```
pdeperio: Shouldn't you keep this uncommented in case of using 'exponential' ```pe_pulse_model``` below?

requirements.txt
```diff
@@ -2,7 +2,7 @@ wheel>=0.23.0
 numpy>=1.10.2
 pymongo>=3.0
 matplotlib>=1.3.0
-scipy>=0.15
+scipy>=0.17
```
pdeperio: Our common installations are at ```0.18``` already (was upgraded due to some crashes before, not sure if ```0.17``` is sufficient).

JelleAalbers:
Thanks Patrick.

The plot shows the median PMT waveform around hits found in single electrons (I selected high-amplitude hits to get a better resolution in the tail). So:
  * I look for single electrons, then take the hits that constitute them
  * For each hit, I take the PMT waveform the digitizer recorded around that hit (unless there are multiple hits per pulse, then I skip it)
  * I align these waveforms based on their maximum amplitude
  * At each sample, I compute the median amplitude over all these waveforms. That's shown in the plot.

SE pulses almost always contain one pe per pulse. Even if a double hit occurs, and even if the explicit rejection of double-hit pulses doesn't removed it (because the hits are merged), taking the median should prevent e.g. afterpulses from having a noticeable effect.

There is other evidence that such a late tail is present:
  * You can see it in single photoelectron waveforms (see e.g. Shingo's notes: [here](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:analysis:kazama:gain_calib_for_analysis) and [here](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:analysis:kazama:pmt_calibration:fitting_method)).
  * In the LED gain calibration, a long integration length (>10 samples) has to be used to get a gain that isn't biased by the integration window (see one of Shingo's notes, not sure which is the best reference). This must mean there is a significant late area component.
  * The mean area per single-pe hit is currently about 0.75 pe, whether you look at lone hits, LED data, single electron or S1 hits. This indicates the current hitfinder is missing 1/4 of the pe pulse area; this motivated #479.

You can also take lone-hit waveforms to show this effect. I've done this here: https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:aalbers:double_pe_emission#updatesource_of_the_discrepancy_in_mean_gain
which shows the effect is present in every PMT.

pax/config/XENON1T.ini
JelleAalbers: Yes, it's better to keep the rise time and fall time uncommented so people can switch by changing just one setting, I'll change it.

requirements.txt
JelleAalbers: I'll bump to 0.18, this has been out for a while, and we don't know if it works on stuff we don't use ourselves

PR #479, opened 2017-01-16
Hitfinder upgrade to accomodate long-tailed pe pulses

JelleAalbers:
This splits the hitfinder in three separate plugins, and changes/simplifies the integration region determination. The main motivation is the recent realization that the long PMT pulse tails (see #478, #369, #363) should not be ignored, but integrated, else we get nonlinearities in our energy scale (since in a large signal they are always integrated due to many hits piling up). 

The code split is motivated mainly to improve maintainability: a large monolithic hitfinder is very intimidating to change or, for new people, to get into. Now it is split into three parts:
  * `PulseProperties.PulseProperties`, which computes the baseline and noise level in each pulse.
  * `HitFinder.FindHits`, which finds the hits
  * `HitfinderDiagnosticPlots.HitfinderDiagnosticPlots`, which does nothing by default, but can make per-PMT hitfinder diagnostic plots. 

The new integration region determination is much simpler; in principle, we just integrate a fixed time before and after each hit (defined as an interval above threshold), with a few details to prevent double counting (hits are first found simply as intervals above threshold, then extended; left extension has priority over right extension). Here are some examples of the new hitfinder in action (on simulated data with the new waveform models introduced in  #478 and the noise bank from #471):

event0000_pulse16456-16557_ch037

event0000_pulse10454-10605_ch037

event0000_pulse08590-08693_ch028

Notice the second and third example both have two very closely packed hits (the second also has a much further-away hit of course); in one case they are merged into a 2pe hit, in the other they are split. The current hitfinder would certainly have merged both into one hit, since it required a drop below the low threshold (a concept which now no longer exists) to separate hits.

There is no more dynamic threshold coefficient. This might cause some interesting changes for very high energy events, where we get very long (several us) tails.

Other changes:
  * The baseline is no longer computed as max(average pre-trigger, average post-trigger), but simply as the average pre-trigger, as it was in the past. I introduced the more complicated method because I thought the long-pe-pulse tails were something we could and should ignore (but now, as mentioned above, I realize we have to integrate them).
  * The hitfinder threshold is set to a minimum of 15 ADC counts, the self-trigger threshold. If we don't do this, we might get a very small extra amount of very small hits in channels where the noise level is very low -- but only when there is a neighbouring larger hit. We don't want the hitfinder efficiency to depend on the presence of neighbouring hits; that  would cause a distortion of the PMT hitpatterns.
  * The height/noise threshold is raised from 4 to 6. Based on some preliminary studies by @arjenwildeboer, this should have very little impact; most hits below 6 h/n are already removed by the self-trigger threshold (since the noise level is generally 2 ADC counts or so). 
  * The hitfinder threshold computed per pulse is stored with each pulse for convenience (so the analyst / diagnostic plotting plugin doesn't have to recompute it). Since we hardly ever save pulses in the processed data (only for S1s) the impact on data size should be minimal.
  * Support for the MessagePack output format. (see #455).
  * Option to shrink raw data by preprocessing baselines, then cutting pre- and post-selftrigger regions from pulses for high energy events (see #455). This is not enabled. Use with caution, as we now know our photoelectron pulses have long tails, cutting away the post-trigger region will bias the area of isolated single-pe hits downwards. For the large S1 and S2 this does not matter much, but for the single electron tail it does.

I would not recommend merging / reviewing this in detail just yet:
  * We need a few performance studies with simulated data (see #478), to check the integration lengths set here are reasonable. @arjenwildeboer is working on this.
  * Look at a couple of real waveforms of different energies (very low to very high) to verify things are not going crazy.
  * Think about the consequences of negative-area hits, which are now possible (though very unlikely). Would this crash pax?
  * Try this out on a larger dataset to ensure pax does not crash.


JelleAalbers:
I had a look at a few 100 PMT waveforms; it seems to be behaving as expected. The worst things I could find were some cases where a lot of noise gets integrated, e.g.

event0006_pulse100631-100734_ch141

This is inherent in the algorithm. I don't see how we can avoid this without introducing a bias due to the long-tailed PMT pulses.

Another thing that sometimes happens is this:

event0003_pulse100003-100126_ch041

If the tail of a hit is oscillating a bit, it can cross the threshold a few times, creating multiple hits where the should be one (two in this case). For peak area determination this matters little, but perhaps it has an effect on clustering of S1s. It doesn't seem to be very common though.

I made one change (apart from a few fixes to the diagnostic plotting system): hit areas and height can now no longer be negative. These would happen very rarely when a few samples of the integration window are split off from a hit due to the local minimum clustering, which has (and must have) the power to break hits. This would cause the plotting code to cry, and possibly other downstream code as well. Now areas and heights are capped at 1e-9. So if you see a hit height/area of 1e-9 somewhere, don't panic.

JelleAalbers:
Thanks Arjen! If there are no further comments I will merge this tomorrow.

PR #481, opened 2017-01-31
Add pickle output for plotting waveforms

coderdj:
This is a simple commit that lets you store waveforms as pickles matplotlib figures.

I hope this might be useful in a waveform watching campaign where we could let people peruse waveforms on their own machine (not over X-forwarding, which is slow) without requiring a pax installation or worrying about versioning.

To save you set Plotting.plot_format to 'pickle' and run pax with --plot_to_dir {your_output_dir}

This is all you have to do to load:
```
import pickle
import matplotlib.pyplot as plt
pickle.load(open("yourfile.pickle", "rb"))
plt.show()
```


pax/plugins/io/Plotting.py
```diff
@@ -91,6 +93,9 @@ def finalize_plot(self, event_number=0):
         if self.output_dir:
             if self.config['plot_format'] == 'pdf':
                 plt.savefig(self.output_dir + '/%06d.pdf' % event_number, format='pdf')
+            elif self.config['plot_format'] == 'pickle':
+                ax = plt.gcf()
```
tunnell: __[flake8]__
* [W291] trailing whitespace

PR #482, opened 2017-01-31
Implement (r,z) position correction from Junji's simulation

JelleAalbers:
This includes the (r,z) position correction due to nonhomogeneous fields in pax and the waveform simulator. The implementation in the simulator is needed; otherwise we'd have to specifically turn off the correction during MC processing.

When data is processed with a pax that includes this change, the interaction positions are corrected for you, and you don't need to do anything in analysis.

  * The magnitude of the corrections is stored in new interaction attributes, so you can recover the old positions if desired. This is different than for S1 and S2, whose areas are stored uncorrected, and external code (e.g. hax) multiplies them with the corrections. Doing it that way for (r,z) correction would force some annoying changes downstream (remake minitrees etc). 
  * The corrections are **additive**, not multiplicative as for the S1 and S2 areas (no good reason, but too lazy to change now).
  * The (r,z) correction is applied **before** we call the S1(x,y,z) correction or S1 pattern goodness of fit. Thus, these maps should expect to be given true/corrected coordinates, not distorted coordinates. The effect should be minor for the S1 LCE (which varies quite weakly in r) even though our current map is derived on distorted coordinates. For the S1 pattern fit it should simply be an improvement, as this comes from Monte Carlo.

Correcting positions in old/existing data
===============================
If you want to correct positions processed by older versions of pax, you can use the correction map in your analysis code. Here's a very basic example, showing how to use the inverse correction (which the simulator uses) followed by the actual correction:

    from pax import utils, units
    from pax.InterpolatingMap import InterpolatingMap
    filename = 'transverse_field_rz_correction_12kv.json.gz'

    # Comment this line if you've downloaded the file yourself, rather than using
    # the version that will ship with pax after this PR has been merged.
    filename = utils.data_file_name(filename)

    rzmap = InterpolatingMap(filename)

    r = 42
    z = -80
    print("True position: r=%0.1f, z=%0.1f cm" % (r, z))

    r += rzmap.get_value(r, z, map_name='to_distorted_r')
    z += rzmap.get_value(r, z, map_name='to_distorted_z')
    print("Observed position: r=%0.1f, z=%0.1f cm" % (r, z))

    r += rzmap.get_value(r, z, map_name='to_true_r')
    z += rzmap.get_value(r, z, map_name='to_true_z')
    print("Corrected position: r=%0.2f, z=%0.2f cm" % (r, z))
    
This gives:

    True position: r=42.0, z=-80.0 cm
    Observed position: r=39.8, z=-80.3 cm
    Corrected position: r=41.99, z=-79.94 cm

You can get started right away if you download 'transverse_field_rz_correction_12kv.json.gz' from this branch and comment the indicated line above. Or, you can wait until this is merged and deployed, then it will work out of the box. 

Intermezzo: nice figure
==================
Here's the distortion as applied by the simulator:

distortion

Details on the correction
===================

The correction map is derived from Junji's simulation (xenon:xenon1t:junji:comsol:dir4:overalloct15). He included quite a lot of points, which I downsampled a bit. The code to convert his text file to the pax correction map format is here:
https://github.com/XENON1T/XeAnalysisScripts/blob/master/mapmaking/Parse%20Junji's%20textfile%20for%20(r%2Cz)%20correction.ipynb

I tested the implementation in pax works  by simulating a 1000-electron S2 at (x=44cm, y=0, z=-90 cm):

    instruction,recoil_type,x,y,depth,s1_photons,s2_electrons,t
    0,NR,44,0,90,1000,1000,0

Pax reconstructes the peak at (x=41.5, y=-0.1), showing the simulated distortion works. The main interaction, however, has  the correct position at (x=44.05, y=0-0.12), thanks to the correction.


JelleAalbers:
Feel free to review this, but if there are no objections I will merge it tomorrow. We can of course update the map later.

PR #483, opened 2017-02-01
Saturation correction

feigaodm:
This branch features a waveformed based sateration correction for S2 signals above 1e4PE. The correction significantly improves the position reconstruction at high energies, and leaves the low-E signals for WIMPs search unaffected. More details about the correction can be found in this note (https://xecluster.lngs.infn.it/dokuwiki/lib/exe/fetch.php?media=xenon:feigao:xenon1t_background_comparison_jan2017.html).

Here are some bref description of the key features.
1， New WF shape model is determined basd on all non-saturated PMT channels, this WF shape model is used to determine the S2 width and integration edges
2,  S2 signal for saturated channels are re-calculated based on the WF model. For non-saturated channels, the area is also updated according to the new integration window
3,  Position reconstruction is performed based on the corrected area per channel.

### DIFF
--- a/pax/config/XENON1T.ini
+++ b/pax/config/XENON1T.ini
@@ -41,10 +41,11 @@ dsp = [
 compute_properties = [
                         # Classify the clusters based on the properties
                         'ClassifyPeaks.AdHocClassification1T',
+                        
                         # Saturation correction
-                        #'PeakAreaCorrections.S2SaturationCorrectionWF',	
+                        'PeakAreaCorrections.S2SaturationCorrectionByWF',
+                        
                         # Position reconstruction
-
                         'WeightedSum.PosRecWeightedSum',
                         'MaxPMT.PosRecMaxPMT',
                         'RobustWeightedMean.PosRecRobustWeightedMean',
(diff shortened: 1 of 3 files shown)


JelleAalbers:
Hi Fei and Yuehan,

Thanks a lot for implementing this! I think it's a great idea to derive the waveform shape from non-saturated channels, and use this to correct the per-channel waveforms. That means we're independent from any optical or waveform shape model, and that things other than area (such as widths, area fraction top, etc) can also be corrected.

Looking at the code makes it clear to me what you want to do now. I think it's probably easier to implement this at a very early stage in pax; even before the hitfinder. Then you don't have to worry about recomputing any hit/peak properties; now you have to duplicate some functionality of earlier parts of pax to do this (either copying the code, or doing something that's hopefully equivalent but may not be).

I can also see you like programming in C :-) In python, if you write e.g. for loops over numpy arrays, the resulting code becomes a bit slow. Numpy has a variety of tricks (mostly under the name of vectorization) that allow you to do some of the things you want to do faster (using less typing and less CPU cycles).

I took the liberty of making a "remix" of your algorithm, implemented at an earlier stage in pax. Maybe you can have a look at that to see if you like it (and/or suggest improvements)? Alternatively we could try to work together to solve some issues with the code that you have (code duplication, speed, some minor things) but I think that might take a bit longer / lead to a more complicated result.

Jelle


p.s.
You say it's only for S2s, but I see that your code also operates on S1s, and it seems to improve them as well.

JelleAalbers:
Remixed in #484

PR #484, opened 2017-02-02
"Remix" of Fei&Yuehuan's saturation correction

JelleAalbers:
This is a "remix" of Fei&Yuehuan's saturation correction (#483), implemented at an earlier stage in pax. This means we can do it a little bit faster, and don't have to recompute any peak properties.

As also explained in [Fei & Yuehuan's note](https://xecluster.lngs.infn.it/dokuwiki/lib/exe/fetch.php?media=xenon:feigao:xenon1t_background_comparison_jan2017.html) the idea is :
- Derive the waveform shape from non-saturated channels;
- Scale this down to an individual channel using some 'reference samples' close to the saturated region;
- Use the scaled waveform model to estimate what happened beyond the digitizer dynamic range.

Here the region 10 samples before and after a saturated region is used for computing the scale (but the length is configurable). If there are multiple saturated regions per pulse, the non-saturated samples in between the saturated regions are also used. Here's an example:

desat_principle

(where 'other PMTs' means: other PMTs that didn't saturate in this region). I tested this on a simple simulated S1 + S2 waveform where both have a medium amount of saturation:

As you can see, I turned off all PMT afterpulses and single electrons to get a good MC/data matching (since the afterpulses / single electrons are not recorded in the truth information, it will be poor otherwise). Here are the results:

|                                                     | Time taken (ms) | Extra time (ms) | S1 area | Error | S2 area | Error  |
|-----------------------------------------------------|-----------------|-----------------|---------|-------|---------|--------|
| Truth                                               |                 |                 | 34543   |       | 2053483 |        |
| No desaturation                                     | 270             |                 | 31897   | 7.7%  | 1799325 | 12.4%  |
| Hitpattern-based desaturation (#237)              | 270             |                 | 34559   | 0.0%  | 2063888 | 0.5%   |
| Late pulse-based desaturation (#483)          | 780             | 510             | 33617   | 2.7%  | 0       | ?      |
| Early pulse-based desaturation (here) | 400             | 130             | 34867   | 0.9%  | 2049454 | 0.2%   |

While #237 performs very well, it assumes the MC LCE per PMT map (which is exactly true in this simulated event, but probably not yet in the real data) and can't correct other properties of the peak (width, area fraction top, etc). There is probably a small bug in #483 that caused the S2 area to be set to 0 just for this example. 

This code is still a bit slow, because the sum waveform in other channels is recomputed for every saturated pulse. If there are many pulses that saturate at the same time (which is usually the case) there is probably a better way to do it. Fortunately not every event saturates, and real events have to spend a lot more time wading through the single electron tail after an S2, so the impact on pax's speed should not be so much.


feigaodm:
Hi Jelle, 

Thanks for the modification! I believe you have a better idea where this de-saturation should go. Here are some questions I'd like you to clarify (cross-check).
1, the peak edges need to to more accurately determined. But I didn't see how you re-defined the peak edges based on the non-saturated summed wf. This must be done for several reasons. A, we only would like to scale signals that is coherent in all channels, not things like PMT after-pulses that apper only in a particular channel. B, More correct edges have a very significant impact on position reconstruction, consider that the hit pattern of Photon-ionization S2s are not as localized as real S2s and we have to remove them when calculating the peak area per channel.
2, It's much better to only use the WF samples before the saturation as the reference samples for scaling. I had check it that using both left (rising) and right (dropping) edges yields much worse saturation correction.  
3, Are you using all non-saturated samples as references for scaling, a little confused to see the marks on the WF you shows
4, The speed is mostly limited by calculating the summed WF. I checked that the SumWF module in pax takes similarly amount of time, maybe you can calculate SumWF in this module and do not repeat the calculation later to save time.

I will take a more carefull look in the afternoon, but your plot looks very reasonable to me!

Thanks,
Fei

JelleAalbers:
Hi Fei,

Thanks!

1. The concept of 'peaks' does not exist at this early stage in pax (not even hits). You're right, however, that high-energy S2s are usually grouped together with part of their single electron tail at the moment; if we want to change this we should look at the clustering plugins. On A, do you think that's a possibility for a physical signal from the TPC? For weird nonphysical things I'm not sure it matters what we do (as long as we don't make them smaller, which I explictly forbid in the code). We could refuse to change the waveform if the ratio saturated ch / others is too large.

2. I assume that's because of the photoionization tail? We could do as you suggest, though I suppose because I use only 10 samples on either side of the saturated region the effect is limited (I would guess the S2 or S2 is still dominant in the first 100 ns after the saturation). 

3. No, only the 'reference samples' incicated in red. Outside these we i

4. It's tricky because here I need the sum waveform only for the nonsaturated channels. I suppose we could try computing the sum waveform for all channels first, then subtracting the sum waveform for saturated channels... At the moment though, summing the waveform has to happen after the hitfinder, as we also compute a hits-only sum waveform (which is in fact used for things like computing widths).

feigaodm:
Hi Jelle,

Thanks!

Here are some of my considerations!

1,  I think there is probably disagreement between our analysis targets using this de-saturation technique. A very important target in this module is to improve the energy resolution, hence the statistical uncertainties of the correction must be as small as possible. With that saying, we have to use as much reference samples as possible instead of picking some samples. 

2, In the simulation, we didn't simulate based saturation effects, hence you can use the tail (drop edge) for the scaling. But in real data, I saw there are some channels propably suffering the base saturation as well. So the tail of that signal is not scalable. So I would only use the rising edge for the correction.

3, The other reason why I don't want to use droping edge is to get ride of much contribution from photon-ionisation. This will affect the position reconstruction. 

4, For the same reason, as I mentioned, we have to define the peak edges more precisely for both saturation correction and position correction. In this sence, maybe we should move the saturation correction after we have a "peak" concept in pax.

 5, there could be some potential bugs in the original desataturation code, but most of the steps are needed to achive good position and energy reconstruction.
 
Thanks again!

JelleAalbers:
I think we're on the same page: the goal is to reduce the bias due to ADC saturation and the loss of resolution that occurs because of it. 

Good points about using the rising edge only. I've changed it as you suggested.

I now increased the number of reference samples to 100 (1 us before the saturated region). Let me know if you'd like to have a different default; we can always change it later. I'm not sure if using many samples will help or hurt: this is all at high amplitude, so statistical fluctuations are small; on the other hand, the further away we go from the saturated region the less correlated we can expect the signal to be, so systematic effects could come in. Perhaps this is worth studying at some point.

Do you think we should apply some further condition to avoid working on non-coherent signals or signals where everything but a few channels is saturated, e.g. only apply the correction if the area in the other channels is at least X% of the total?

The peak edges are indeed not very good at high energy at the moment, but that's the clustering algorithm's job to fix; I don't it has to do with saturation. I'm having a look at the clustering now/soon, I think there are things we can improve at high and low energy.

I'll rerun the test above when I am back in the office to check I didn't just introduce an awful bug. Just let me know or just commit to this branch if you'd like to change something.

feigaodm:
Thanks for the changes! Just a few more remarks:)

I think/hope the desaturation is also usefull to get rid of base saturation. 

For the number of reference samples, I believe the larger the better (of course if it's still valid). I don't understand why you have to fix it. The best way is to keep it floating according to the saturation condition. Sometimes there could be only 20 samples that are not saturated and sometimes there could be 200. So I would insist to keep it floating.

The idea of juding the coherentness of a signal in a particular channel is very good, and it's on the to-do list to further improve the performance of the correction. 

The peak edges is essential to improve the performance further from my experience, and we can tune it further. I agree with you that we can probably do it in the hit clustering plugin. But from my experience of processing WFs, it's easy to define the peak edges based on the summed WFs. I don't know how easy it is to do it in clustering plugin.

Lastly, why not we test your branch on real data as we did? Besides the improvement on position reconstruction, would you please produce such an anticorrelation plot to verify the improvement on energy resolution?

test_saturation

If we have similar, or even better performance than the one as is, I'm fine to merge it and do tuning later.

Thanks, Fei

JelleAalbers:
Thanks Fei. There were a few minor bugs in my last commit, but they are fixed now. I agree we should test on data and tune further based on that.

Just for reference, here's how the reference region looks now:


desat_final_fornow


|                                                     | Time taken (ms) | Extra time (ms) | S1 area | Error | S2 area | Error  |
|-----------------------------------------------------|-----------------|-----------------|---------|-------|---------|--------|
| Truth                                               |                 |                 | 34543   |       | 2053483 |        |
| Early pulse-based desaturation (old) | 400             | 130             | 34867   | 0.9%  | 2049454 | 0.2%   |
| Early pulse-based desaturation (now) | similar             | 130             | 34636   | 0.3%  | 2056245 | 0.3%   |

PR #487, opened 2017-02-08
Add Adam's S2(x,y) correction map

JelleAalbers:
This adds Adam's S2(x,y) correction map derived from Krypton data. The map is described in his notes:

https://xecluster.lngs.infn.it/dokuwiki/lib/exe/fetch.php?media=xenon:xenon1t:adam:fv-study:s2-correction-xy-kr83m-json.html

Thanks very much to Adam for providing this map in the complete pax-compatible format on such a short notice. :-)

I made a few small changes to the map to make it compatible with the earlier correction conventions:
  * I inluded the analytic part of the correction (two_dim_quad) into the map values. Although this is slightly less accurate, it means we don't need to adjust any code in pax, and can  use the same plugin as is used for XENON100's (x,y) correction. Perhaps we should upsample the map the next time we do this (or just bite the bullet and code a new plugin). 
  * I normalized the map so the geometric average of the map is 1. That is, a uniformly distributed monochromatic peak (like Kr83m) will not shift in S2.
  * The map is now a relative light yield map, as pax wants to divide by the correction map value.

The code I used for the normalization is here: https://github.com/XENON1T/XeAnalysisScripts/blob/master/mapmaking/Normalize%20Adam's%20S2(x%2Cy)%20LCE%20map.ipynb

The final map looks like this:
adam_map

Dots are the data points of the map, between which we interpolate.

Adam has also made an S2(x,y) correction for the top and bottom array separately. Currently we don't have variables in the pax datastructure for this, but we can easily add them in the next release. We'll have to remake the map soon anyway when the Kr data has been reprocessed (new hitfinder, field correction, eventually also new gain model).

@adambrown1 do you agree with these changes or would you like to see something different?


adambrown1:
Those changes are all fine; next time I produce the map I can just do those few things (and upsample) before saving it. You can 'see' the loss of accuracy (or lack of) by comparing to the plot below where I apply the analytic part separately - I think that final map looks OK.

S2 map

Although there are no variables for 'cs2_bottom' there is 's2_area_fraction_top' so it is possible for analysers to extract those values (and indeed this is done for some analyses). That said, it is no problem to wait until another release to add this support.

JelleAalbers:
Thanks Adam! Yes, let's put the top and bottom correction in next time.

PR #490, opened 2017-02-09
Fix for negative samples

ErikHogenbirk:
This branch introduces a bugfix for events with a large negative area in a hit. As the hitfinder now has no low threshold, but just appends samples, some of the samples can be negative. For the total area, this is already fixed. However, the center of gravity can still be skewed so that it is outside the pulse. This is fixed here by excluding the negative part of the hit-only sum waveform for the center of gravity computation.
Second, there is a minor update for the number of hits that can be found in one pulse.


ErikHogenbirk:
No problem, if you want me to take even more credit for other work that you've done just let me know ;)

Issue #491, opened 2017-02-09
S2 bottom pattern hit map not used in fax & s2_mean_area_fraction_top not compatible with the map used


mcfatelin:
After a more close look at code:

- I found that currently the S2 bottom pattern hit map is not used in simulating the S2 hits. See [L637 of pax.simulation](https://github.com/XENON1T/pax/blob/master/pax/simulation.py#L637). The bottom S2 pattern is assumed to be uniform.
- In the mean time, fax uses a constant value of s2-top-fraction to allocate photons for a simulated S2 on top/bottom PMTs (See [L245 of XENON1T.ini](https://github.com/XENON1T/pax/blob/master/pax/config/XENON1T.ini#L245) ). And the value is not updated to the latest value got from new MC simulation (See [here](https://github.com/XENON1T/XeAnalysisScripts/blob/master/mapmaking/S2%20LCE%20mapmaking.ipynb) ). This is the reason we see discrepancy of AFT between data and MC in [Joey's note Fig.3 ](https://xecluster.lngs.infn.it/dokuwiki/lib/exe/fetch.php?media=xenon:xenon1t:sim:notes:jhowlett:preliminary_fax_only_studies_170131.html)

Suggest to:

- Update the s2_mean_area_fraction_top of [L245 of XENON1T.ini](https://github.com/XENON1T/pax/blob/master/pax/config/XENON1T.ini#L245) to the latest value from [MC simulation](https://github.com/XENON1T/XeAnalysisScripts/blob/master/mapmaking/S2%20LCE%20mapmaking.ipynb)
- Or in long term to include also bottom PMTs in S2 hit pattern map so that we get rid of the separate treat of S2 hit on top/bottom PMTs. [L637 of pax.simulation](https://github.com/XENON1T/pax/blob/master/pax/simulation.py#L637)

JelleAalbers:
Yes, this is correct. I agree changing the value in the ini would be the easy solution for now, including the bottom PMT hit map would be the best solution long-term.

JelleAalbers:
I'll change the title since I understand you've resolved the second part of the issue. Once a top+bottom S2 map is made, we'll still have to change fax to support it (and still support top-only maps ideally).

l-althueser:
Fyi: The latest S2 map is extracted from MC for top+bottom PMTs. Changes in PAX/FAX are still missing.

PR #501, opened 2017-02-14
Fax update for campaign

mcfatelin:
New update of fax specifically for the use of the analysis campaign. The main changes are:

1. PMT gains and gain sigmas are now from PMT WG, which accommodate the condition in the first science run.
2. Change the s2_area_mean_top_fraction to the updated value in MC. See [Issue#491](https://github.com/XENON1T/pax/issues/491)
3. Corrected the PMT after pulse probability. Previous one was 100 times higher.
4. The time resolution of ion-residual PMT after pulse is added. Values are from JulienW
5. Also the uniformly time-distributed PMT after pulse was implemented. See [summary note](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:analysis:firstresults:waveform_simulator#pmt_afterpulses) for more details of such type of PMT after pulses. Because it can happen very close to the primary hit, it is potentially a important responsible for peak finding/classification/clustering efficiency loss.

However this will not be the last update of fax during the analysis campaign. The foreseen update on fax would be the implementation of multi-PE PMT after pulse generation.


JelleAalbers:
Merged in other pull request

PR #504, opened 2017-02-15
Cluster based on central regions of hits

JelleAalbers:
Since the recent hitfinder upgrade (#479), hits are much broader. This means there are fewer gaps between the hits. Several clustering algorithms look for such gaps in their clustering, and therefore perform worse now.

This partially alleviates the problem by clustering on the central regions of each hit -- the regions that actually went above the hitfinder threshold (usually 15 ADC counts). These are stored in the new left_central and right_central attributes of the Hit class.

This doesn't fix the problem completely: for example, closely packed PMT signals often end up in a single hit, which can't get split by the natural breaks clustering (LocalMinimum can, but only at high energy). I have some ideas to solve this (in the asym_splits branch), but they are pretty big changes that could make things worse in unexpected ways. I'd rather not do these things unless there is an indication there is a problem with the current clustering settings.


pax/dsputils.py
```diff
@@ -95,31 +95,56 @@ def get_detector_by_channel(config):
             detector_by_channel[ch] = name
     return detector_by_channel
 
+@numba.jit(numba.void(numba.float64[:], numba.int64[:, :], numba.int64, numba.int64),
```
tunnell: __[flake8]__
* [E302] expected 2 blank lines, found 1

PR #507, opened 2017-02-16
Update SE gain, add double-pe emission

JelleAalbers:
This adds a double photoelectron emission probability of 0.15 to fax. It is not applied for afterpulses (because they are not made by UV light, and because it is easier to code this way :-)

I also updated the single electron gain to 23 photons/single electron, see https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:aalbers:single_electron_gain_stability. We can probably derive a more accurate value later.


pax/config/XENON1T.ini
```diff
@@ -244,6 +244,9 @@ s1_patterns_file =                    'XENON1T_s1_xyz_patterns_Feb2017.json.gz'
 s2_patterns_file =                    'XENON1T_s2_xy_patterns_top_updatedgeo.json.gz'
 s2_mean_area_fraction_top =           0.56             # See XeAnalysisScripts/mapmaking
 
+# Probability of oduble photoelectron emission
```
pdeperio: typo

JelleAalbers: Nah oduble photoelectron emission is a special magic process, don't you know ;-)

pax/simulation.py
```diff
@@ -257,6 +257,13 @@ def make_pax_event(self):
 
             photon_detection_times = np.array(photon_detection_times)
 
+            # Add double photoelectron emission
+            if len(photon_detection_times):
+                n_dpe = np.random.binomial(len(photon_detection_times), p=self.config['p_double_pe_emision'])
+                if n_dpe:
+                    dpe_times = np.random.choice(photon_detection_times, n_dpe, replace=False)
+                    photon_detection_times = np.concatenate([photon_detection_times, dpe_times])
```
pdeperio: Easy to add extra truth info for DPE?

JelleAalbers: Not from this place in the code, though I agree it would be useful to store somewhere

PR #510, opened 2017-02-17
Add more shape properties, update classification and interaction pairing

JelleAalbers:
This adds two additional shape properties we expect to be useful in classification:
 - **tight_coincidence**: the number of channels that have a peak maximum within 50 ns (+- 2 samples) of the sum waveform maximum of the peak. This follows the suggestion by @marcschumann to define a tighter coincidence window. Hopefully this will be useful in distinguishing real S1s from any junk.
 - **area_decile_from_midpoint** (array of 11 floats):the time from the area midpoint until the area decile (time where X0% of the peak area is passed) is reached. This is zero for the 50% area point, and negative for earlier points. This follows the idea from @feigaodm to use the rise time as a variable in discriminating S1s from single electrons.

A (rather weak) test on computing the area decile has temporarily been removed to support this. "We" should write a better test in the future.


JelleAalbers:
I added two more changes we discussed: 
  * New classification: peaks with tight_coincidence <= 2 are labeled unknown, rest are S1 or S2 based on rise time.
  * New pairing rule: start with S1 with the largest tight_coincidence, then move down. Degeneracy in tight_coincidence is broken by area (largest->lowest).

PR #524, opened 2017-02-20
Effective S1 shape model

JelleAalbers:
This adds, and activates as default, an option to replace the S1 excimer decay time distributions by a simple exponential. Setting this to 35 ns (instead of the expected 24 ns for triplets and 3 for singlets) produces an S1 shape model which matches the Kr83m data:

s1_model

Just using the literature values for excimer decays does not produce a match:

default_singlet_f

The likely explanation is that the waveform simulator does not account for the photon travel time. This can of course be simulated by our optical monte carlo, and @pdeperio and others are working to include it in fax. 

s1_widths_comp

The S1 widths observed in Kr data show some larger outliers than in the simulation (consistent with this hypothesis, since propagation time has large tails). The rise time (which we use for classification) should be less affected.


JelleAalbers:
Just checked and the S1 model is still matching.

PR #526, opened 2017-02-22
Explicit type in neural net

tunnell:
This is due to new numpy requirement.  Can't take float anymore.


tunnell:
@mcfatelin this should fix your problem

PR #534, opened 2017-03-01
S1 LCE Map for SR0

berget2:
Updated S1 LCE map with November Kr83m data, processed with pax_v6.4.2, using FDC from https://github.com/XENON1T/pax/pull/512

Should be updated in conjunction with FDC


JelleAalbers:
Is it ok to merge this so we can use it for re-correction, or would you prefer to wait to do it at the same time as the new field correction map?

berget2:
The maps are well labeled, so we can just add more as we go with different FDC designations. However if you'd prefer to wait that's fine too.

JelleAalbers:
Ok, thanks, I have no objection (this adds a file we're going to need anyway and may already be useful)

PR #537, opened 2017-03-02
Simulation: Use special adc thresholds, settings for pre-filterbox

JelleAalbers:
This fixes a bug in the self-trigger emulation configuration that cause the special (per-PMT) thresholds to not be picked up. 

This also adds a configuration to simulate in conditions appropriate for our before-filterbox runs (such as AmBe), with higher self-trigger thresholds and a different noise sample extracted from a noise run close to the AmBe.


JelleAalbers:
@mcfatelin any objections?

PR #538, opened 2017-03-06
Field Distortion Correction Update

berget2:
FDC updated with Junji's r-only correction.

Produced and examined with https://github.com/XENON1T/XeAnalysisScripts/blob/master/mapmaking/Parse%20Junji's%20textfile%20for%20(r%2Cz)%20correction.ipynb


JelleAalbers:
Superseded by #551

PR #551, opened 2017-03-10
Berget2 lce fdc update

berget2:
Added newest FDC and S1 LCE corrections


JelleAalbers:
Thank Ted! I'll merge this so we can see if it works as we expect.

Issue #553, opened 2017-03-15
n must be >= x (in S1AreaFractionTopProbability.py)


pdeperio:
@ershockley encounters the following error processing ```XENON1T-7862-000097000-000097999-000001000.zip```, event 212 of this zip (97212 of run):
```
Traceback (most recent call last):
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.6.1_OSG/bin/cax-process", line 11, in <module>
    load_entry_point('cax==5.0.5', 'console_scripts', 'cax-process')()
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.6.1_OSG/lib/python3.4/site-packages/cax-5.0.5-py3.4.egg/cax/tasks/pro\
cess.py", line 359, in main
    _process(*sys.argv[1:])
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.6.1_OSG/lib/python3.4/site-packages/cax-5.0.5-py3.4.egg/cax/tasks/pro\
cess.py", line 159, in _process
    core.Processor(**pax_kwargs).run()
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.6.1_OSG/lib/python3.4/site-packages/pax-6.6.1-py3.4.egg/pax/core.py",\
 line 315, in run
    self.process_event(event)
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.6.1_OSG/lib/python3.4/site-packages/pax-6.6.1-py3.4.egg/pax/core.py",\
 line 276, in process_event
    event = plugin.process_event(event)
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.6.1_OSG/lib/python3.4/site-packages/pax-6.6.1-py3.4.egg/pax/plugin.py\
", line 91, in process_event
    event = self._process_event(event)
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.6.1_OSG/lib/python3.4/site-packages/pax-6.6.1-py3.4.egg/pax/plugin.py\
", line 108, in _process_event
    return self.transform_event(event)
[...]

PR #555, opened 2017-03-15
Fix peak.hits_fraction_top calculation

pdeperio:
Should fix https://github.com/XENON1T/pax/issues/553

Was a bug in calculating peak.hits_fraction_top. @darrylmasson S1AreaFractionTop cut in lax should probably be restudied.


tunnell:
Why the change?

Issue #556, opened 2017-03-17
binom_test MemoryError


pdeperio:
Discovered by @lucrlom processing ```170316_0210``` (event number?):
```
Traceback (most recent call last):
  File "/project/lgrandi/anaconda3/envs/pax_v6.6.2/bin/cax", line 9, in <module>
    load_entry_point('cax==5.0.12', 'console_scripts', 'cax')()
  File "/project/lgrandi/anaconda3/envs/pax_v6.6.2/lib/python3.4/site-packages/cax-5.0.12-py3.4.egg/cax/main.py", line 142, in main
    task.go(args.run)
  File "/project/lgrandi/anaconda3/envs/pax_v6.6.2/lib/python3.4/site-packages/cax-5.0.12-py3.4.egg/cax/task.py", line 65, in go
    self.each_run()
  File "/project/lgrandi/anaconda3/envs/pax_v6.6.2/lib/python3.4/site-packages/cax-5.0.12-py3.4.egg/cax/tasks/process.py", line 217, in each_run
    ncpus)
  File "/project/lgrandi/anaconda3/envs/pax_v6.6.2/lib/python3.4/site-packages/cax-5.0.12-py3.4.egg/cax/tasks/process.py", line 104, in _process
    core.Processor(**pax_kwargs).run()
  File "/project/lgrandi/anaconda3/envs/pax_v6.6.2/lib/python3.4/site-packages/pax-6.6.2-py3.4.egg/pax/core.py", line 315, in run
    self.process_event(event)
  File "/project/lgrandi/anaconda3/envs/pax_v6.6.2/lib/python3.4/site-packages/pax-6.6.2-py3.4.egg/pax/core.py", line 276, in process_event
    event = plugin.process_event(event)
  File "/project/lgrandi/anaconda3/envs/pax_v6.6.2/lib/python3.4/site-packages/pax-6.6.2-py3.4.egg/pax/plugin.py", line 91, in process_event
    event = self._process_event(event)
  File "/project/lgrandi/anaconda3/envs/pax_v6.6.2/lib/python3.4/site-packages/pax-6.6.2-py3.4.egg/pax/plugin.py", line 108, in _process_event
    return self.transform_event(event)
  File "/project/lgrandi/anaconda3/envs/pax_v6.6.2/lib/python3.4/site-packages/pax-6.6.2-py3.4.egg/pax/plugins/interaction_processing/S1AreaFractionTopProbability.py", line 34, in transform_event
[...]

PR #558, opened 2017-03-29
Skip p-value calculation for large s1

darrylmasson:
Fix for issue #556


pax/plugins/interaction_processing/S1AreaFractionTopProbability.py
```diff
@@ -14,13 +14,16 @@ def startup(self):
...
 
     def transform_event(self, event):
 
         for ia in event.interactions:
             s1 = event.peaks[ia.s1]
 
-            if s1.area < self.low_pe_threshold:
+            if s1.area > self.max_s1_area:
+                continue
+            elif s1.area < self.low_pe_threshold:
```
JelleAalbers: You don't need the second condition here due to the continue (and if you do want to have it, `else` would work just as well)

JelleAalbers: The second condition can be if instead of elif (if the first condition is true, the continue makes sure you don't reach the second)

darrylmasson: You're right, the elif isn't giving us anything. Old habits.

darrylmasson:
@pdeperio `ia.s1_area_fraction_top_probability` is initialized to `float('nan')`, but I'm not sure how many of the "AllEnergy" cuts in lax really work above 1e4 pe.
@JelleAalbers I think what's causing the memory issues it it's making arrays of approximate length `s1.area`, so capping it at 1e4 should use negligible memory. If we discover it's still a problem I'll hopefully have rewritten the functions to not need these large arrays.

PR #562, opened 2017-04-19
Coherent noise and maximum S2 afterpulse delay

JelleAalbers:
This adds two options to the waveform simulator, which are by default turned off:
  * An optional **maximum delay time for S2 afterpulses**. When testing the peakfinder's efficiency, packing several S1s into one simulated event gives a great speedup. However, S2 afterpulses extend to a maximum drift time after the initial signal, so you'd have to space the peaks very far apart-- nullifying the speed increase (since pax must simulate noise everywhere in between). This adds an option to remove all S2 afterpulses that occur after a certain time. This is only for peakfinder efficiency studies, you don't want to enable this if you simulate full events. 
  * An optional **coherent noise simulation mode**. As explained in earlier pull requests (#152, #471), pax pulls noise in small (480 samples) samples from a randomly triggered 'event'. To get enough variation out of the limited noise sample included with pax, each channel draws independent samples. This adds an option to instead draw samples once per event, so the noise is coherent across channels. The 480-sample samples are still drawn randomly in time from the template, so don't expect your simulated noise to have the right Fourier transform.

Note that you have much less variation in the noise if you use the coherent noise simulation mode, unless you supply a much larger noise template file yourself.


pax/simulation.py
```diff
@@ -453,21 +463,30 @@ def make_pax_event(self):
...
+                                                                    needed_noise_samples)
+
+                    roll_number = np.random.randint(noise_sample_size)
+
                 # Extract the chosen noise samples and concatenate them
                 # Have to use a listcomp here, unless you know a way to select multiple slices in numpy?
                 #  -- yeah making an index list with np.arange would work, but honestly??
                 real_noise = np.concatenate([
-                    self.noise_data[channel - self.channel_offset][nsn * sample_size:(nsn + 1) * sample_size]
+                    self.noise_data[channel - self.channel_offset][nsn * noise_sample_size:(nsn + 1) * noise_sample_size]
```
tunnell: __[flake8]__
* [E501] line too long (121 > 120 characters)

JelleAalbers:
I added:
  * **Noise rolling**, the drawn noise samples are shifted right by a random amount (up to 1 noise sample), to avoid artifacts in the noise model falling in the same few places every time, especially in coherent noise simulation;
  * **Remove afterpulse code duplication**; the gain sampling code for afterpulses was duplicated, and I ran into some issue with it when I added a new dead PMT without updating the afterpulse parameters. @mcfatelin, could you have a look to see if the functionality you put in is preserved?

tunnell:
Can this go in?  Seems build is failing

mcfatelin:
@JelleAalbers Your changes about AP seems okay to me. Actually the previous AP was still using truncated Gaussian. Now you update it to adapt also the realistic SPE shape. Thanks.

JelleAalbers:
@mcfatelin Thanks! 
@tunnell I hope I'll fix it soon, though there's no rush to include this in the next reprocessing as it only affects the waveform simulator.

JelleAalbers:
Travis says it's fixed now

PR #563, opened 2017-04-19
FDC & LCE SR1 Update

berget2:
Added first SR1 FDC and LCE maps, and updated default corrections in XENON1T.ini accordingly


pdeperio:
this one failed checks but you still merged?

PR #565, opened 2017-04-19
Add science run 2 S2 map

adambrown1:
Adds new S2 maps computed using the science run 1 Kr83m calibration data.

See xenon1t:analysis:firstresults:s2-xy-correction-summary for a summary of the SR0 map, the new map is produced in the same way.


tunnell:
Add a description?  I mean what are you doing here?  How was map made?  etc

coderdj:
The top comment has a link to a wiki note (here: xenon1t:analysis:firstresults:s2-xy-correction-summary) that describes the method. Or where do you want further explanation @tunnell ?

sanderbreur:
For future us: This is for science run 1, not science run 2 ;)

s

Op ma 24 apr. 2017 om 18:59 schreef Christopher Tunnell <
notifications@github.com>:

PR #566, opened 2017-04-21
Add base saturation region reference treshold

sanderbreur:
This improves: "Remix" of Fei&Yuehuan's saturation correction #484

#484 was originally meant to desaturate S2 signal to improve the data vs MC spectrum comparison. As it was implemented at a level where it could not know if the peak was an S1 or S2 it now also desaturated S1 signal. For S1 signals this method was over correcting the signals by up to a factor 2. The reason for this over correction was that in the saturated S1 channels the number of reference samples above 1PE is ~2, which is not enough (rise time is to fast) to get a correct ratio. (for S2 channels this is approx 30 to 80 ref samples) To solve this we now check how many reference samples we have before deciding to correct the individual channel. 

This pull request will now keep the S2 desaturation intact, but makes sure it doesn't desaturate S1 signals. This effects mostly the alpha analysis, where now it was impossible to use any saturated S1 signals. (approx everything below -50 cm)

Before #484 The alpha singlas looked like this:

before_desat

After #484 we see that the S1 signals below ~50 cm are over corrected:

after_desat

When looking at individual waveforms we see the following:

170418_saturation_s1s2_waveforms

(I have looked at a few hundred waveforms which show mostly the same) The S1 gets over corrected because there are one a few reference samples above 1PE and before the saturation happens. For S2 signals, which are much broader, there are enough reference samples to make a good correction. (this example waveform of the S2 was one of the worst)

So for several events I checked how many reference samples are used. For this I selected high energy gammas (we expect the S2 to be saturated) and high energy alphas (we expect the S1 to be saturated):

reference_samples alpha_gamma

So I tested a threshold of a minimum of 10 reference samples above 1PE to desaturate. I tested this by manually processing a selection of 1000 high energy gamma events and 300 alpha events. The goal is to make sure the cS2 is still the same as before this pull request, and that the S1 is now not desaturated.

First look at the alphas where we expect the cS1 to now have much lower values for the regions where saturation happens:

effect_on_alphas_in_s1

effect_on_alphas_in_s2

Now lets look at the gamma events which should not have changed in S2:

effect_on_gammas_in_s1

effect_on_gammas_in_s2

The conclusion is that alpha S1 signals are now not being desaturated and the S2 signals of gammas are still desaturated, which was the goal. (A next improvement will be to actually correctly desaturate S1 signals, but for now this works. and in the future there are some small improvements to be made on the S2 desaturation)


pax/plugins/signal_processing/DesaturatePulses.py
```diff
@@ -66,6 +66,11 @@ def transform_event(self, event):
                 # The pulse is saturated, but insufficient information is available in the other channels
                 # to reliably reconstruct it
                 continue
+
```
tunnell: __[flake8]__

pax/plugins/signal_processing/DesaturatePulses.py
```diff
@@ -66,6 +66,11 @@ def transform_event(self, event):
                 # The pulse is saturated, but insufficient information is available in the other channels
                 # to reliably reconstruct it
                 continue
+                
+            if len(w[reference_slice][w[reference_slice]>1]) < self.config['reference_region_samples_treshold']:
```
tunnell: __[flake8]__
* [E225] missing whitespace around operator

pax/plugins/signal_processing/DesaturatePulses.py
```diff
@@ -67,6 +67,11 @@ def transform_event(self, event):
                 # to reliably reconstruct it
                 continue
 
+            if len(w[reference_slice][w[reference_slice]>1]) < self.config['reference_region_samples_treshold']:
```
tunnell: __[flake8]__
* [E225] missing whitespace around operator

feigaodm:
Thanks Sander! Looks good to merge. For the S1 desaturation, I think we should use the hit pattern based desaturation you developed.

PR #568, opened 2017-04-30
Prepare for supernova trigger

JelleAalbers:
This modifies the eventbuilder for implementation of a "dead man's switch" for preserving the untriggered data if there is a galactic supernova:
  * Acquisition monitor pulses are now rescued (written to .pickles file) even if delete_data is False;
  * The eventbuilder has a new mode activated by including the --deleter argument. In this mode it will not act as a normal event builder, but look for runs with the trigger.status field set to `deleteme`, then delete their untriggered data.

To complete the supernova switch, we'd need to add a program that sets trigger.status to `deleteme` if the trigger status for a run is already at 'processed' and there was no supernova during that period.

Then there would be a few minor further changes required:
  * Change the run mode configuration to include delete_data, except in background data-taking. We then no longer start the eventbuilder with --delete_data (which overrides what's specified in the run config).
  * Change the website status display to no longer proclaim imminent death and destruction when there is a couple of GB in the trigger buffers.

As of now this is **completely untested and will not work yet**. I hope that somewhere in the coming weeks there will be a convenient moment where I could deploy and test this PR without too much downtime and/or shifters being scared by the inevitable DAQ errors, deleted data, etc.


JelleAalbers:
Please ignore the codacy comment, it seems to think using defensive programming with 'assert' is a code quality issue...

coderdj:
I started to look through this. What happens if we're running in a fast mode (say Rn220) where we want to delete still within the run and not just entire runs at once. Is there a flag to set that keeps the current configuration or allows deletion within a run?

Sorry if this is obvious from the code. It's not completely easy to follow jumping in for the first time.

JelleAalbers:
Hi Dan, Thanks for looking! Yes, if we put `delete_data = True` in the MongoDB options for pax, we will keep running in the current configuration. One of the steps towards deploying this would be to add this option to the default run configuration, and override it with `delete_data = False` for background_stable (or other low-rate modes).

coderdj:
Bump. What's needed in this? Need a review? Testing?

JelleAalbers:
I added a script, `delete_decider`, which:
  * Sets the delete permissions for runs. The code watching for new supernova triggers should go in here. Currently it just marks a run for deletion if the trigger is finished on it and data acquisition ended more than 2 hours ago. Note that for some runs untriggered data is already deleted during event building (calibration data). 
  * Allows us to add a manual external trigger on the current run (for testing);
  * Print a status table of which runs are currently marked for preservation;
  * Should be runnable outside the DAQ network (if you set the hostname for the runs db correctly). 

Note that this is separate from the script that actually issues the delete orders to MongoDB; that is done by `event-builder --deleter`. This has to run on the event builder machines, as it has to access the untriggered db.

Looking forward to deploying this tomorrow and Friday, if all goes well... @feigaodm, after this is deployed to the DAQ (and the inevitable bugs fixed), can we merge this into master? It is convenient to have the DAQ run on the master branch of pax.

JelleAalbers:
Deployment successful, I'm merging so the DAQ can stay on the master branch.

feigaodm:
Thanks Jelle. I was slower to merge it. Hopefully we will be able to see supernovas soon:)

PR #570, opened 2017-05-02
New S2 map

adambrown1:
This adds a new S2 correction map (v2.1) for science run 1 and sets it as the default in XENON1T.ini.

It replaces an earlier map (v2.0) which had been produced using data processed with the incorrect gains, which was pull request #565.


coderdj:
We need to move these corrections files out of pax so we don't have to keep bumping pax version just to update minitrees. But since this is not totally simple let's just merge this for now and kick the can down the road a bit.

PR #572, opened 2017-05-12
Patch for S1 area fraction top plugin

darrylmasson:
Fixes some lingering issues with the S1 area fraction top probability plugin, including excessive memory usage at large S1s (which caused the failure of MC jobs), and also extends the functions to accept floating point inputs, which improves accuracy at low energies.

See note https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:darryl:s1_aft_update

Tested in fax up to 8e5 photons with no problems found. Performance in [0,100] keV is ~1.3 ms/event, in [100,10000] keV ~3.5 ms/event.


pdeperio:
thanks darryl.  did you check if this also fixes the supposed leak over time?

darrylmasson:
I'm seeing slightly decreased memory usage, but not significantly. I'm running another test at constant energy but it'll take more time to finish.

image

darrylmasson:
So for simulated events of 100 keV I see ram usage increase 13.8 MB/min with this patch and 15.1 MB/min without. It's an improvement but probably not the culprit.

pdeperio:
ok thanks for checking.  are you able to isolate if this leak is in the fax (simulation) stage or pax (reconstruction) stage or both?  (See [this example](https://github.com/XENON1T/processing/blob/master/montecarlo/run_sim.sh#L310) of splitting fax and pax stages.)

darrylmasson:
pax vs fax memory usage:
pax_fax_memory
These were the same jobs (ie ran fax first, then ran pax on the output in separate instances), each 1000 events at 100 keV from the same fax instruction set. fax memory usage climbed at 16.7 +/- 1.0 MB/minute, pax memory usage flat.

PR #576, opened 2017-05-25
classification update

feigaodm:
Mis-identification of single electron S2 signals as S1 signal is a big contribution for the Lone-S1 rate in the detector, and significantly affect the [accidental coincidence background](https://xecluster.lngs.infn.it/dokuwiki/lib/exe/fetch.php?media=xenon:xenon1t:analysis:firstresults:ac_model_background_0425.html).

[A dedicated cut ](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:analysis:firstresults:exploring_se_cut) was defined by @MAVJ to reduce the probability of mis-identification. Instead of putting this cut in lax stage, we'd like to modify the peak classification in pax stage instead, which motivates this PR.

A minor modification of original cut is done to have slightly better acceptance for high AFT S1s (events happening on the top part of the TPC). Efficiency of this cut should be estimated together with other pax efficiency loss, such as [this note](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:sim:notes:jhowlett:efficiency_summary_170227)


pdeperio:
can you include more documentation (note or plot) showing the tuning you did here?

MAVJ:
pax665_clasif_limit

That's how the boundary looks in PaxV6.6.5

pdeperio:
This should be compared to "Cut v4_RvsAft" in Figure 2 of [your note](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:analysis:firstresults:exploring_se_cut) (which was used in [SR0 lax](https://github.com/XENON1T/lax/blob/v0.11.1/lax/lichens/sciencerun0.py#L687))?  But then they look very similar, i.e. shouldn't the curve drop from AFT 0.4 to 0.5 according to this PR?
```
self.s1_rise_time_aft = interpolate.interp1d([0, 0.3, 0.4, 0.5, 0.70, 0.70, 1.0],
                                             [70, 70, 65, 60, 35, 0, 0], kind='linear')
```
Can you overlay the two?

And then how about a plot of the change in ```s1_rise_time_bound```?

pdeperio:
@feigaodm @MAVJ ping!

@jhowl01 just pointed out that this has [changed again in pax_v6.8.0](https://github.com/XENON1T/pax/pull/596/files#diff-4efccdcceb8434c9708c1e9e65a0182dL8).  Can we please get more documentation on this, i.e. the overlay plot requested above?

PR #488, opened 2017-02-08
New per-PMT LCE maps for full LXe+GXe volume

pdeperio:
@l-althueser and I prepared these new MC per-PMT LCE maps with [MC v0.1.7](https://github.com/XENON1T/mc/releases), including [his optical parameter tuning to Kr83m S1 data](https://xecluster.lngs.infn.it/dokuwiki/lib/exe/fetch.php?media=xenon:xenon1t:sim:notes:althueser:opphotstudy_kr83m_nov16.html#aa-s1-matching), and the full LXe+GXe volume from bottom to top PMT faces. 

There is an extra R bin above z=0 (gate) to cover the small region shown in [Fig. 3e here](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:fvm:xe1t_newxenon1tdesign_svn_r300#pictures_from_mc_meshlab_identification_of_the_volumes). Below the gate, this extra R bin is set to the same LCE as the previous R bin as a stopgap solution to an undesirable binning artifact in the pax conversion notebook below.

The maps are derived using the analysis framework [here](https://github.com/XENON1T/MCAnalysisScripts/tree/master/PerPMT_LCE_Time) following a similar binning as @berget2 (described in [my note](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:sim:light:pdeperio:lce_timing)), and converted to pax format using [this notebook from @JelleAalbers ](https://github.com/XENON1T/MCAnalysisScripts/blob/master/PerPMT_LCE_Time/Parse_for_pax.ipynb), including a few cross-checks therein.

Future work includes:
- Developing an external program to derive and compare different maps 
- Implementing the option to use this single map for everything, including S1(x,y,z) LCE correction currently coming directly from @berget2's data map

PR #577, opened 2017-05-28
New per-PMT LCE maps for full LXe+GXe volume (mc v0.2.1 and pax v6.6.5)

l-althueser:
I have prepared a new MC per-PMT LCE map for the full LXe+GXe volume with [mc v0.2.1](https://github.com/XENON1T/mc/releases) which includes a new optical parameter setup.

Details on the optical parameter setup can be found in the corresponding [mc fitting to data note](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:sim:notes:althueser:xe1t_201705_mc_matching). This is the first time we can use one parameter setting for S1 and S2 simulations. The map is still named "s1" but could also be used for s2 generation.

The map is created with the tools delivered by @pdeperio and @JelleAalbers and should be equal to what was implemented in #488. A detailed notebook and a comparison of the old and the new map can be found here: [mc map comparison](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:sim:notes:althueser:xe1t_201705_perpmt_lce)

Do I have to change the map name on more places as I found?

PS: For GitHub it seems that I have renamed the file but you can trust me that it is a new map ;)


pdeperio:
thanks for the very complete documentation!  

looks like the difference is very small, maybe just a little less at the top of the overall LCE; could you include a difference plot as well?

l-althueser:
The difference is less than i thought two days before (as I used two different binnings ..). I have added some difference plots in the note.

pax/config/SimulationMCInput.ini
```diff
@@ -1,7 +1,7 @@
 [pax]
 parent_configuration = "Simulation"
 
-s1_patterns_file =                    'XENON1T_s1_xyz_patterns_Feb2017.json.gz'
+s1_patterns_file =                    'XENON1T_s1_xyz_patterns_May2017.json.gz'
```
pdeperio: You can remove this line now (inherit from XENON1T as default)

PR #582, opened 2017-06-03
InterpolatingMap.get_value should not accept kwargs

JelleAalbers:
This fixes a bug in InterpolatingMap.get_value, which accidentally accepted all keyword arguments. It should only accept the `map_name` argument. As it is now, people won't get an error if they misspell the argument (e.g. `map` or `map_naem`), instead, the default map (`map`) will silenty be used instead of the map the user intended to specity.


feigaodm:
any further considerations/checks? @JelleAalbers

JelleAalbers:
Sorry, the build is failing because I'm using a python-3 only construct. I will try to fix it later (but it's not very urgent I guess)

JelleAalbers:
If you're putting *coordinates after map_name, you can't call it unless you specify a map_name; might as well make map_name mandatory. I think code in hax and notebooks may be using this function, so if you change how it's called we'd have to modify that too.

The other solution is let it accept arbitrary **kwargs at first, but check kwargs for the presence of anything other than map_name (and error if that is the case). In fact this is why I let it accept arbitrary kwargs originally (but I didn't consider the problem of argument checking)

feigaodm:
Thanks Jelle! Yeah, there is some problem when we used in some other parts of pax. I guess I will leave it as is for now. Maybe you can take a look at it when you are free.

PR #589, opened 2017-06-27
FAX bug fixed. The PI is now with drift process simulated as well.

mcfatelin:
In the previous fax, the electron loss during drifting is simulated, but only for the main S2s. It is not done for the photoionization (PI) after pulses. In principle, the PI rate is also related to the electron loss during drifting (higher the electron lifetime, the higher the PI rate). 

In this PR, the electron loss due to drifting is added for the PIs.

Before:
sim_rn

After:

updated_sim_rn copy

(plot credit: Alexander Shaw, summer student at Columbia University)


pax/plugins/io/WaveformSimulator.py
```diff
@@ -211,15 +211,23 @@ def s2_after_pulses(self, g4_id=-1):
...
             # Delete afterpulses coming after the maximum delay time (if this option has been enabled)
             delay = delay[delay < self.config.get('max_s2_afterpulse_delay', float('inf'))]
-            num_s2_afterpulses = len(delay)
 
+            # Simulate the electron loss
+            rvs_uniform = np.random.uniform(low=0, high=1., size=len(delay))
+            delay_after_drift = []
+            for dt, rv in zip(delay, rvs_uniform):
+                surviving_prob = np.exp(-dt/self.config['electron_lifetime_liquid'])
+                if rv<surviving_prob:
```
tunnell: __[flake8]__
* [E225] missing whitespace around operator

PR #592, opened 2017-06-28
TopPatternFunctionFit

pelssers:
To test and compare the TopPatternFit improvement TopPatternFunctionFit it is added to pax. This way we can directly compare the performance of the two methods after partial reprocessing. For a full reprocessing I'd recommend to choose either TPF xor TPFF since having both plugins slows down pax.

This adds TopPatternFunctionFit as a posrec plugin to Pax. It's implemented as a Derived class of TopPatternFit since it works exactly the same except for the PatternFitter instance which is changed in startup() to load the fitted LCE maps instead of the LCE maps TopPatternFit uses.

The XENON1T.ini config is changed to include TPFF as the last posrec plugin to run.

A unittest for TopPatternFunctionFit is also added.


pelssers:
Watching the Travis builds I don't see TopPatternFunctionFit listed in the Pax Timing report for the various tests, however I did enable the plugin in XENON1T.ini, is this correct or does it need to be added in another place as well?

PR #598, opened 2017-08-11
Updated map for s1 area fraction top

darrylmasson:
A finer binning means better ability to work with regions where area_fraction_top changes quickly, like near the edges of the detector.


darrylmasson:
The stuff I presented this morning was run through this map - half a million events in [1.5, 15000] PE

PR #600, opened 2017-08-22
Don't keep LED data in buffer

JelleAalbers:
The LED trigger config is hardcoded in the eventbuilder, not specified by the run doc. After the changes to how we run the eventbuilder following #568, LED data is no longer automatically deleted, but only after some time (just like background data). Obviously we're not going to detect supernovae while we're triggering on our LED pulser. This changes the defaults so data is also online-deleted in LED mode.


JelleAalbers:
Indeed, it does not affect data processing. There might have been a DAQ issue this week that this pull request will resolve, so if possible I'd like to deploy this later this week (after operations' approval of course).

Issue #599, opened 2017-08-21
Fraction not reached in waveform


ershockley:
In pax 6.8.0, while processing Rn220 runs there was a single zip file from one run that has the following error. This is from Run 11269, zip file XENON1T-11269-000007000-000007999-000001000.zip, event 67. 

Not sure yet how common this is - it occurred in 1 run out of the recent reprocessing of 50 Rn runs. 

```
Traceback (most recent call last):
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.8.0_OSG/bin/cax-process", line 11, in <module>
    load_entry_point('cax==5.0.5', 'console_scripts', 'cax-process')()
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.8.0_OSG/lib/python3.4/site-packages/cax-5.0.5-py3.4.egg/cax/tasks/process.py", line 455, in main
    _process(*sys.argv[1:])
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.8.0_OSG/lib/python3.4/site-packages/cax-5.0.5-py3.4.egg/cax/tasks/process.py", line 163, in _process
    core.Processor(**pax_kwargs).run()
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.8.0_OSG/lib/python3.4/site-packages/pax-6.8.0-py3.4.egg/pax/core.py", line 315, in run
    self.process_event(event)
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.8.0_OSG/lib/python3.4/site-packages/pax-6.8.0-py3.4.egg/pax/core.py", line 276, in process_event
    event = plugin.process_event(event)
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.8.0_OSG/lib/python3.4/site-packages/pax-6.8.0-py3.4.egg/pax/plugin.py", line 91, in process_event
    event = self._process_event(event)
  File "/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.8.0_OSG/lib/python3.4/site-packages/pax-6.8.0-py3.4.egg/pax/plugin.py", line 108, in _process_event
    return self.transform_event(event)
[...]

PR #601, opened 2017-08-22
fix pax bug for very rare event

feigaodm:
This PR tries to fix a bug reported by @ershockley when processing Rn220 calibration data with pax_v6.8.0. This bug is due to de-saturation plugin as this event has very wired saturation. Since not a lot channel are detecting signals during saturation, we should skip this event for de-saturation correction. This will fix issue  (#599).

Waveform of this events is here.
stupid_waveform


sanderbreur:
After checking it I think we are adding a needed check, but losing another needed check. Because iether the pulse itself or the sumw can have not enough samples. So I would keep this change, but add the deleted part back into the code. (so do it both for w and for sumw)

This should be added back:
-            if len(w[reference_slice][w[reference_slice] > 1]) < self.config['reference_region_samples_treshold']:
 -                # the pulse is saturated, but there are not enough reference samples to get a good ratio
 -                # This actually distinguished between S1 and S2 and will only correct S2 signals
 -                continue

feigaodm:
Thanks @sanderbreur , just added it back and merged the PR.

sanderbreur:
Perfect, ty

Op wo 6 sep. 2017 om 00:21 schreef Fei Gao <notifications@github.com>:

PR #606, opened 2017-09-05
PMT gains in fax updated for SR1

mcfatelin:
A quick PR for updating the PMT masks/gains used in fax. The gain values are from [Shingo's note](https://xe1t-wiki.lngs.infn.it/lib/exe/fetch.php?media=xenon:gain_sr1_20170815.html)


pdeperio:
How can we easily switch between different configs, e.g. SR0 vs SR1?

mcfatelin:
Maybe in the next PR where we update more about fax for SR1, we need to create an separate ini file for SR0 in case people want to use. But in principle it is better to have ini file connected to database.

pdeperio:
Soon we will consider to re-analyze SR0, including re-calculation of all fax deliverables.  So would be nice to have in this PR an input file with config option similarly to [other maps that already have run dependence](https://github.com/XENON1T/hax/blob/master/hax/treemakers/corrections.py#L73).

pdeperio:
For future reference, I guess this list was taken at time ```test_t = 1502383532``` (GMT: Thursday, August 10, 2017 4:45:32 PM).

We may wish to later check the effect of using gains from beginning of run versus end of run as a systematic.

pdeperio:
Two ```.ini```s for toggling implemented in https://github.com/XENON1T/processing/pull/43

Issue #607, opened 2017-09-14
Zero sized aquisition monitor files


XeBoris:
We observe that there are zero-sized acquisition monitor files (*pickles) produces by pax on xe1t-datamanager when e.g. a noise dataset is taken or an LED calibration is done. These datasets cause some trouble in the rucio catalog.
Rucio distributes the raw data (files attached to a rucio container) according to transfer rules. Once the transfer is finished the raw dataset (a rucio container) is marked from transferring to "ok" which means that all single files are transferred. Unfortunately this does not work when there is a zero sized file in the raw data. Some rucio storage elements (RSE) do not allow this. The status of the rucio catalog entry switches to "stuck" after a while and the dataset is not transferred correctly and therefore it does not appear in our runDB!
My main question if it possible to prevent pax from creating a zero sized file like this? The only alternative is to delete zero sized files before uploading to the rucio catalog (in ruciax). Or is this file needed for anything, nevertheless if it has a size or not?

Thank you,

feigaodm:
fixed by #608

PR #608, opened 2017-09-14
Add random initial pickle to acquisition monitor file

JelleAalbers:
This addresses #607 by adding some random nonsense at the start of each acquisition monitor pickle file.  There will be a companion pull request in hax to make the acquisition monitor pulse reading code robust to this.


JelleAalbers:
There will be an event builder update later this week for #600, would you like this to go along as well?

PR #611, opened 2017-09-18
New FDC map using data-driven method

jingqiangye:
Data-driven 3D position correction in SR1, see note here:
https://xe1t-wiki.lngs.infn.it/doku.php?id=jingqiang:xenon1t_kr_fdc


pdeperio:
@coderdj @ershockley I have manually (begrudgingly) added ```FDC_SR1_data_driven_3d_correction_v0.json.gz``` to the ```pax_v6.8.0``` environment on Midway, so we can continue specifying v6.8.0 in the Midway triggered jobs (v6.8.1 may have some other small changes that make it incompatible with v6.8.0).  Please double check and run some test jobs of course.

PR #612, opened 2017-09-27
New SR0 Net for SR0 reprocessing

weiyuehuan:
In SR0 campaign,  the parameters for S2 simulation are not optimised. After SR0, Lutz optimised the simulation parameters by matching MC and Kr83m data. After that, we validated the new simulation with SR0 data:
1) https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:sr1:positon_reconstruction
2) https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:yuehuan:pax_versions_wall_leakage
SR1 neural net was trained based on the new simulation(mc_v0.3.0).  **For SR0 reprocessing, we also need to updated the net for it.**
**To  short:**
1) NN_XENON1T_v8_mc_v030_sr1_n15.npz   already using the optimised map. For SR1.
2) **This PR upload NN_XENON1T_v8_mc_v030_sr0_n15.npz for SR0**

Be noted: 
1) we use different net for SR0 and SR1 due to PMT configuration changed.
2) all the outdate nets are deleted to avoid confusion in future.


pdeperio:
@coderdj @ershockley I have manually (begrudgingly) added ```NN_XENON1T_v8_mc_v030_sr0_n15.npz``` to the ```pax_v6.8.0*``` environments on Midway and CVMFS, so we can continue processing with v6.8.0 (v6.8.1 may have some other small changes that make it incompatible with v6.8.0).  Please double check.

@coderdj Then I guess you'll just need to update the corrections DB and RunsDB with this new NN file for SR0 runs, and reprocessing should be able to restart without changing anything else.

coderdj:
@pdeperio ya update to runs DB is ready to go. We just have to make sure this file available (as you did). I'm not well versed enough in the GRID stuff to know if adding to the environment is sufficient to make GRID processing work but @ershockley will know.

pdeperio:
I've also manually deleted the old NN versions from all ```pax_v6.8.0*``` environments as an extra precaution.

ershockley:
Will that affect daily processing?

pdeperio:
No, daily processing should be using ```NN_XENON1T_v8_mc_v030_sr1_n15.npz```, which hasn't changed.

PR #614, opened 2017-10-05
Fix last pulse computation in apparently not impossible edge case

JelleAalbers:
There appears to be a very strange edge case where it seems a collection contains data, but a moment later it seems to be empty. I've apparently seen this before or at least anticipated it (see https://github.com/XENON1T/pax/blob/master/pax/plugins/io/MongoDB.py#L277).

The trigger should always keep well away from the last point where data is inserted. It does this by looking at the last pulse in the last collection that has data, then subtracting one minute for safety (except at the end of the run of course). When the aforementioned edge case occurs, the trigger instead computes the last time from the collection start time. Unfortunately there is a unit conversion error in this piece of very rarely used code (factor 10 for Mongo samples vs the internal pax unit of nanoseconds). 

As a result, when the weird edge case occurs, the trigger thinks data exists until kingdom cometh and proceeds to query as far ahead as it can. Since it only fires off a limited amount of queries, it gets about one minute ahead, until the next check for data makes it come to its senses. The result is a small (~one minute) region of emptiness in the data, possibly preceded by a small region with partial events (with data from only a few digitizer). 

Acquisition monitor data saved in the pickle files is still ok, that's saved by different logic.

We'll have to remove these regions from the data, as partial events can cause anomalous leakage (though it's usually very obvious from e.g. the S2 hitpattern when you have such an event). There are several ways you could detect this problem,
  * a gap in the pulse counts of the trigger monitor
  * a warning in eventbuilder.log:
```
10-05 07:59 MongoDBReadUntriggered INFO     Last subcollection with data is 98
10-05 07:59 MongoDBReadUntriggered WARNING  Latest subcollection 98 seems empty now, but wasn't before... Race condition/edge case in mongodb, bug in clearing code, or something else weird? Investigate if this occurs often!!
10-05 07:59 MongoDBReadUntriggered INFO     Submitting query for subcollection 95
10-05 07:59 MongoDBReadUntriggered INFO     Submitting query for subcollection 96
10-05 07:59 MongoDBReadUntriggered INFO     Submitting query for subcollection 97
```
  * a discrepancy in the dead time reported by the trigger and that deduced from the aqm busy_on and busy_off pulses saved in the pickle file, e.g. :
deadtime_discrepancy
  * (probably more things, e.g. last_pulse_time info in the pipeline_status collection)

This fixes the unit conversion error, so should prevent the trigger from running ahead. I have no idea what causes the edge case itself; maybe something in MongoDB?


JelleAalbers:
I'm merging to deploy this on the DAQ

feigaodm:
Looks like there is some issue in the Travis test. See https://travis-ci.org/XENON1T/pax/builds/283710483?utm_source=github_status&utm_medium=notification

JelleAalbers:
It's an issue with all the builds at the moment, something weird with apple?

JelleAalbers:
For python 2 it looks like it's trying to compile numpy from scratch, since the numpy that `conda` gets (1.9.something) is out of date.

For python 3 I don't know what's going on, maybe it installs the wrong version of numpy?

You could try waiting a few days to see if it's a transient bug in the dependency tree that will get resolved, otherwise I suppose permuting the numpy/scipy versions in the `conda create` might help.

feigaodm:
Thanks Jelle!

PR #615, opened 2017-10-05
tuning clustering for SR1

feigaodm:
When evaluating pax efficiency to find S1 and S2s, Joey found that pax performance is not perfect at lowest energy, see Fig 8 of [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:sciencerun1:daqtriggerpaxefficiency#pax_efficiency). By examining lots of simulated WFs, we conclude that there are three directions pax can further improve the ability to merge single electrons from S2s deep in the TPC.

1, NaturalBreaks clustering threshold is too low. See this plot based on FAX simulations. 
screen shot 2017-10-05 at 3 23 25 pm
The original threshold to split the peak into too is too low. By simulation we see that the more than 1% of deep S2s can be splitted as two peaks. So we propose to increase the threshold to a solid line where we expect to improve pax's ability to merge S2 signals. This will probably affect our ability to distinguish multiple scatters, however, as this clustering is applied after the gap_size clustering, the impact of it is much smaller compared to gap_size clustering. This Fix the example WF in joey's note, Fig 10.

2, The gap size clustering threshold is too small (currently set as 2.5us). As shown in the WF below, at a smaller probability ~1%, gaps between deep S2s can be larger than 2.5 us, this is probably due to a larger diffusion constant at a lower field in SR1.
screen shot 2017-10-03 at 3 12 57 pm
So we'd like to fix this by increasing the gap size to 3.0us, by doing this, we are able to merge the S2s again and do not split deep S2s.
screen shot 2017-10-03 at 3 11 32 pm

3, After the first gap_size clustering to isolate S1 signals,  sometimes single electron S2s from deep TPC can be splitted into several S2 clusters. There is a small probability that one (or more) single electron S2s will be mis-classified as S1 signal, and those hits will be removed when we construct S2s. See this WF as an example. 
screen shot 2017-10-05 at 3 38 48 pm
In order to solve this problem, we apply almost the same classification algrithm here during the gap_size cluster algorithms. Then this problem is solved, as seen in the following WF. 
screen shot 2017-10-05 at 3 41 04 pm

Problem 2 and problem 3 both cause ~1% efficiency loss to reconstruct S2s. While we noticed that this effect is small in acceptance, but could be important in terms of ER band leakage as real S2s are splitted.


jhowl01:
I checked the acceptance of small S2s with this branch compared with several earlier pax versions. It seems to mostly compensate the problems introduced by our weaker field. After several iterations of Fei tuning and me rechecking, we decided that we shouldn't keep pushing the max_gap_size_in_cluster parameter to boost the acceptance ad-infinitum, since we may get worse at resolving S2s from multiple scatters.

screen shot 2017-10-17 at 11 53 14 am

For now we decided to merge and let MC people check the resolving power.

PR #616, opened 2017-10-06
Single electron pulse shape update for SR1

katrinamiller:
Update of the anode field domination distance, gas gap warping map, average single electron gain, & triplet lifetime in the gas for science run 1. 

see note: 
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:kmiller:sr1_se_waveform_shape


tunnell:
Need to move gas_gap_warping_map_July17.pkl into data directory

feigaodm:
Thanks @katrinamiller for the update and @tunnell for approval. Will merge it shortly. Then I guess @jhowl01 will need to update S2 efficinecy.

pdeperio:
Please also make a partner PR for the config [here](https://github.com/XENON1T/processing/tree/master/montecarlo/config), which is used in general production.

tunnell:
No no not merge yet

tunnell:
Numpy might be busted, restarted build, might need to pin numpy version

katrinamiller:
Warping map used for SR0 (October 2016):
width_and_gasgap_length_oct16

Warping map used for SR1 (July 2017): 
width_and_gasgap_length_july17

The most recent gas gap warping map was created using data from science run 1 that was processed with the latest pax version (6.8.0). The SE width as a function of x/y position has changed likely due to a difference in the liquid levels between SR0 and SR1.

PR #617, opened 2017-10-06
Fix build dependencies

tunnell:


tunnell:
This branch is intended for fixing some TravisCI issues that revolve around our dependencies.  I think it's particularly an issue with the complex dependency chain that we have due to ROOT, though it appears to arise since Conda isn't getting the latest numpy version?  This is my playing branch.

PR #618, opened 2017-10-07
Travis issue fix

feigaodm:
This is a temporary fix to Travis build issue. Right now python 2.7 is disabled, we can enable it later when matplotlib and/or scipy is fixed. At least we will be able to see whether people's PR can pass the various automatic tests we have.


feigaodm:
@tunnell Done!

PR #620, opened 2017-10-09
remove (old) fdc from pax stage

feigaodm:
This PR removes older version of FDC in pax stage, as it's coved by hax already. Corrections minitrees using TPF should be modified after this change.


feigaodm:
Thanks @pdeperio , corresponding hax PR is made, please check

Issue #604, opened 2017-08-29
Remove Deprecated Data Files


pdeperio:
Can we remove old input data files that should never be used again?  (They should be stored in the version control history anyway, right?)  For example, to avoid issues such as https://github.com/XENON1T/hax/pull/140, or past confusions when running reconstruction with wrong maps.

tunnell:
Of course, delete away!

pdeperio:
@weiyuehuan cleaned up the NN files in #612 (thanks!)

Can the others responsible please clean up their maps?

```s1_xyz```: @l-althueser 
```s2_xy```: @pelssers 
```s1_aft```: @darrylmasson 
```s1_xyz```: @berget2 
```s2_xy```: @adambrown1

JelleAalbers:
If you have old data processed with these maps, keeping them around allows you to re-correct. They'll be in the repo somewhere, but it will be harder for most people to do some study comparing pax versions. I don't know if that's worth some confusion potential like https://github.com/XENON1T/hax/pull/140 -- though weren't these map names set by the corrections db now?

pdeperio:
@JelleAalbers, I was also worried about an (undocumented) issue with a previous reprocessing that took an obsolete NN map filename from the outdated RunsDB, and was only discovered after someone analyzed the position distribution.  (The other maps are indeed taken care of in hax now, but as https://github.com/XENON1T/hax/pull/140 shows, we can still make this kind of mistake there.)

I agree it's a bit more difficult to do direct comparisons without all the maps in one package, but I hope it's enough to have them all in the existing conda environments.

PR #623, opened 2017-10-10
Remove old S1 patterns map

l-althueser:
This removes the old S1 patterns map which I forgot to delete the last time ;) (as suggested in #604)

The map which is removed ("...May...") was already derived from SR1 data. There is no special SR0 map at the moment since no matching of the optical parameters is performed with a recent PAX version. I am quite sure that the current map should be good for both, SR0 and SR1.


l-althueser:
Thanks, I added a comment that will lead to the given MC version in which a detailed description is given (see optical setup macro).

feigaodm:
Thanks @l-althueser , Does the new position FDC have any impact on the data/MC matching you achieved?

l-althueser:
I haven't checked but if the LCE/ly map significantly changes, a new matching of the optical parameters and per-PMT map for S1 and S2 is needed.

PR #626, opened 2017-10-17
Add S2 correction map v2.2 for SR1

adambrown1:
This pull request adds a new version of the S2 correction maps. This is based on science run 1 data which has been processed with pax v6.8.0 and the current corrections as of 16th October. This is foreseen to be the final version for science run 1 analysis.

Full details of the map are on the xenon wiki at https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon1t:adam:s2map:science_run_1_summary

The new map for S2_bottom is:
s2_map_final_bottom

and for S2_top:
s2_map_final_top


l-althueser:
I saw that you are talking about PMT effects and that the holes in the S2_top could be due to missing PMTs. Just in case that you didn't saw it before but I reproduced this behaviour in per-PMT LCE MC maps some time ago and came up with this plot:
s2_aft_excldpmts

sanderbreur:
Well that fits very nice. Are the differences we still see between these
two color maps due to different QE/SPE acceptances between the PMTs? Or
should that already be in the MC that you show Lutz?

Op di 17 okt. 2017 om 14:53 schreef Lutz Althüser <notifications@github.com

adambrown1:
Hi Lutz,

Nice plot, I guess you saw this one I made (the same thing using data
instead of LCE maps). Yours shows that we get a similar pattern from the
MC. Did you just sum up the LCE maps of the PMTs except the turned off
ones? Then I will do the same and compare the two, it will be nice if they
agree.

Cheers

Adam

2017-10-17 14:53 GMT+02:00 Lutz Althüser <notifications@github.com>:

l-althueser:
My plot was from [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:sim:notes:althueser:xe1t_msc_mc_perpmt#S2-Per-PMT-maps-for-v0.3.0) and is generated directly from the per-PMT S2 map (which is also included in PAX) excluding these PMTs. So no additional PMT effects like QE, CE, ... are considered.

adambrown1:
@sanderbreur are you comparing the s2 top map and @l-althueser's when you talk about the differences - not sure quite what you mean? If so then we expect some difference because the LCE does not include the other major effect of electrode warping which goes into the S2 correction.

sanderbreur:
Dear Adam, yes I do. what I mean that if you just compare the locations of
dead PMTs with the darkest areas. So I try to not look at the electrode
warping effect. There are some differences that I assumed should be
explained by QE, CE etc of the indiv pmts

Op di 17 okt. 2017 om 15:12 schreef Adam Brown <notifications@github.com>:

pdeperio:
Should this have removed ```s2_xy_map_v2.1.json```?

adambrown1:
I left it so that it would not break anything before hax was also updated.
Probably that was not necessary, anyway I can make a new PR to remove v2.1
now?

2017-10-25 23:21 GMT+02:00 Patrick de Perio <notifications@github.com>:

l-althueser:
`s2_xy_map_v2.1.json` is still the default in PAX. I preparing a PR which removes some maps, so can I remove this version and add `s2_xy_map_v2.2.json` to the ini files?

adambrown1:
Yes this should be fine; the correction is no longer used in pax directly
but in the hax corrections minitree, which is using v2.2

https://github.com/XENON1T/hax/blob/df668b52ff00839085b7beee49e0cae8dbfcb227/hax/hax.ini#L125


2017-12-20 0:00 GMT+01:00 Lutz Althüser <notifications@github.com>:

PR #630, opened 2017-10-19
correct typo in `add_to_z`

pietrodigangi:
correct typo in `add_to_z` assigned value. 
It was `add_to_z = (2.15 + 0.25) * cm` -> correct is `add_to_z = (2.15 + 0.025) * cm`


feigaodm:
For Xe100 the distance between gate and liquid level is 2.5mm. If we want to set gate to 0, we need 2.5 mm instead of 10 times smaller

feigaodm:
Maybe modify the one in Xe1T.ini file?

pietrodigangi:
@feigaodm there was a misunderstanding with Jelle on gitter chat about that value (the typo is actually in the comment above)

pietrodigangi:
@feigaodm for XENON1T, has this z offset to be the same with impose in the Geant4 code to have the gate at 0? I'm not sure how the value in the Xe1T.ini relates to that one in G4

JelleAalbers:
Yes, Pietro spotted the inconsistency between comment and config, and I agreed it was a typo. However it's the comment that was incorrect while the config was ok. Should be fixed now.

feigaodm:
@JelleAalbers Should we change the one in [XENON1T.ini](https://github.com/XENON1T/pax/blob/master/pax/config/XENON1T.ini#L179) to 0? You commented there that we should change it when G4 geometry is updated, and I think it happended long time ago.

JelleAalbers:
Yes, if the MC files now output positions in the common coordinate system, we can change that setting to 0, and remove the override in https://github.com/XENON1T/pax/blob/master/pax/config/SimulationMCInput.ini.

PR #635, opened 2017-10-25
PMT after pulsing updated for SR1. 

mcfatelin:
PMT after pulses are updated for SR1. Also I separate the PMT after pulse configs for SR0 and SR1 in sub-folders. The default one defined in Simulation.ini is for SR1. The SR1 after pulse is from Run12772 which was taken in August, 2017.


pdeperio:
and corresponding commit to master https://github.com/XENON1T/processing/commit/ccdf00853559442e039efc3b2f4d3504b1d3d916 for SR0 lol

PR #639, opened 2017-10-30
Update single-PE and S1 waveform model

JelleAalbers:
This contains the updated/refined single-PE and S1 pulse shape model described in the following notes:

https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:aalbers:spe_waveform_shape
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:aalbers:s1_waveform_shape

As discussed there, the main change from https://github.com/XENON1T/pax/pull/524 is a slightly narrower single-PE pulse and a slightly broader S1 time distribution than we're currently assuming.

It would be useful to compare the S1 widths in Rn220 data with those from the regular MC production after this is merged (or before, though if you want me to do it, I'm not sure when I'll have the time).


feigaodm:
Will this be used for SR1 efficiency simulation? If yes, should be merged right away

PR #642, opened 2017-11-29
tuning gap_size clustering

feigaodm:
The PR is motivated by a problem @jhowl01 found recently. with a probability of 0.1%, S1 signals can be identified in the hits of single electron WFs. An exmaple WF is attached here

problem_wf1

This is due to a tight gap_size clustering to search for S1 like signal, and hence single electron S2s can be splitted.  If this threshold is too high, we'll have less power to split TPC events happening on the top. 

In order to solve this problem, we need to tune the gap size clustering. Previousely, gap size is defined to take into account the signal extension of a particular hit. However, this introduce a few problems as I understand:
1, The fluctuations in SPE WF can lead to fluctuations in gap size, which probably make it more difficult to tune the gap size parameter.
2, The toy simulation we performed to optimize the gap_size parameter, does not take the SPE WF into account. See this plot for eg. 
screen shot 2017-11-28 at 9 07 01 pm

In this PR, we used hits' 'index_of_maximum' variable to clustering the peaks and it's really easy to understand how those two problems are solved. The thresholds are shown in the above plots. The current S1 gap_size is set as 350ns, we can further increase it if we want. This parameter will allow us to split very nearby S1s I believe, which can be checked with Kr83m analysis. 

I checked two different kinds of WFs @jhowl01 simulated.
1, Deep S2s which triggers this PR.
2, Events happening on top part of the detector where S2 and S1 are close. 

By randomly examining ~20 WFs for each categary, I don't see problem at all. But it would be beneficial if @jhowl01 can check the effect using his matching framework. Below is an example showing that the problematic WF is fixed.
screen shot 2017-11-28 at 9 14 05 pm


jhowl01:
To check the success of this change at its intended goal, I checked our acceptance of small S2s (where large gaps are most common) with this, current pax (v6.8.3) and SR1 pax (v.6.8.0). The improvement is clear:

sr1_s2_pr_comparison

The primary remaining acceptance loss is splitting into S2s, with some smaller fraction (0.06%) having an S1 appear. This probability is cut by a factor of three compared to pax_v6.8.3.

The acceptance of signals near the top of the detector is also retained (up to a ~0.25% difference caused by a bug in fax).

jhowl01:
S1 acceptance also unaffected:

checking_clustering_pr_s1_acceptance

PR #644, opened 2017-12-01
configuration of tight coincidence

feigaodm:
@jhowl01 , please checkout this PR to study effect of tight coincidence on acceptance.


feigaodm:
@jhowl01 I made the change as placeholder for later modifications, effect should be identical to default code, can you help to review?

pax/plugins/signal_processing/BuildPeaks.py
```diff
@@ -31,9 +31,6 @@ def startup(self):
         self.first_top_ch = np.min(np.array(self.config['channels_top']))
         self.last_top_ch = np.max(np.array(self.config['channels_top']))
 
-        # tight coincidence
-        self.tight_coincidence_window = self.config.get('tight_coincidence_window', 50) // self.dt
```
jhowl01: Why was this ever here? It seems it was never used?

pax/plugins/peak_processing/ClassifyPeaks.py
```diff
@@ -26,7 +26,7 @@ def transform_event(self, event):
             # classification based on rise_time and aft
             if -peak.area_decile_from_midpoint[1] < self.s1_rise_time_bound(peak.area):
                 # Peak rises fast, could be S1
-                if peak.tight_coincidence <= 1:
+                if peak.tight_coincidence <= 2:
```
jhowl01: Possible to move this number to the default config?

pax/plugins/signal_processing/BuildPeaks.py
feigaodm: I was trying to add this in the classification decision during clustering, but later found out it's not necessary. But forgot to cut it away.

PR #646, opened 2017-12-05
per-PMT S2 LCE map update

pelssers:
This adds the latest per-PMT S2 LCE maps to pax. The maps are generated using XeAnalysisScripts/mapmaking using the results from Lutz' MC v210 geometry updated simulations.

The maps extend about 2mm further beyond the TPC edge. The new maps have a slightly higher LCE at the maxPMT channel, this could further reduce the under densities observed in the TPF spatial distribution.

Some more details can be found here:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:pelssers:lce_map_compare_sr1


l-althueser:
We could call this map `v2.2.0` and release a new MC version soon (I will have an additional PR for the MC release today). Some more comparison plots of the MC data side can be found here: [geometry update](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:sciencerun1:optical_simulations:geometry_update#updates2_per-pmt_maps)

pelssers:
Ok, I've changed the map name to v2.2.0

l-althueser:
I have just now realized that I used a slightly different S2 generation macro. I used for this production the macro file that was included in the MC versions pre 'v2.1.0' (the one that was used for every other S2 map):
```
/xe/gun/center 0. 0. 4.5 mm
/xe/gun/halfz 0.5 mm
/xe/gun/radius 480.  mm
```
Instead of the adjusted one that have (post `v2.1.0`):
```
/xe/gun/center 0. 0. 4.7 mm
/xe/gun/halfz 0.5 mm
/xe/gun/radius 481.25  mm
```

I wouldn't expect any difference from this and will prepare a PR to the MC to fix this.

feigaodm:
I guess @l-althueser  will prepare a new update soon, do we expected to be a independent map for comparion? If yes, will merge shortly.

PR #650, opened 2017-12-17
tensorflow based nn

weiyuehuan:
update tensorflow NN with:
1, latest MC map: mc_v2.1.0
2, consider PMT collection R/Angular dependence.
3, using high static training samples in high R
 
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:sr1:keras_nn#update2017-12-15_after_telecon


tunnell:
@pdeperio So we decided to merge without even a basic description of the network, basic checks of the network's fit, nor the code available for producing it anywhere?

pdeperio:
@weiyuehuan should add everything later. these are just the files for testing, not yet officially called anywhere nor in a release.

Issue #53, opened 2014-10-10
XEDReader continues on to next sub XED file


tunnell:
Can we make it so the XEDReader can continue reading the next XED file?  Looking at the code, it doesn't look like much work:
- Take file initialization out of startup and into a function
- In get_events, at end of file, see if there is another file.

Issue #43, opened 2014-08-28
BuildWaveforms should deal with overhanging pulses


tunnell:
Occurences can extend outside of the event window.  For example, earlier or later.  This happens because for XENON1T we no longer have a concept of an event.  Therefore, we could have digitized pulses where only some of them fall within the event window.

(If this is hard to do in pax, we may be able to do something in the DAQ.  But this computation has slowed down the DAQ in previous iterations)

PR #652, opened 2017-12-19
Add SR-dependent configs from processing

jhowl01:
This moves the SR-dependent configs used for simulation from the processing repository (created in [processing #43](https://github.com/XENON1T/processing/pull/43)) to `pax/config`. This allows it to be used by multiple fax-related frameworks without separate maintenance/degeneracy.

Also addressed in [processing #53](https://github.com/XENON1T/processing/pull/53).


l-althueser:
I have just realized that many of the maps in these config files are already outdated and not replaced by the new ones since everybody adds maps for HAX and don't care enough to update PAX and remove old maps. I will change some of the maps in a new PR.

l-althueser:
Why do we have the `s2_fitted_patterns_file` entry but not the `s1_patterns_file` or `s2_patterns_file` in there?

feigaodm:
Thanks @l-althueser , that will be nice!

PR #657, opened 2017-12-22
Bugfix for S1_AFT calculation

darrylmasson:
I found an esoteric error in the p-value calculation for the S1AreaFractionTopProbability plugin. Here's what was happening:
problem
This is the probability curve for one event from the 7pe notch. 7.6pe total, area_fraction_top = 0, z around -90cm, expected aft ~0.09, p-value calculated at ~1e-5.

**How does the plugin work?** The goal of the plugin is to find the s1_top values that correspond to where the blue line intersects the black curve. It knows one of them to start with (the left one, in this case), and to look for the other one it starts from the red line (which is mean value of `n*p`) and goes right. As we can clearly see, if you go right from the red line you never reach the blue line, so the successive approximation loop never converges.

**Why was this happening?** The binomial distribution is only defined for integers, and while it's not difficult to extend the definition of the probability mass function to work for floats, the definition of the mean (`floor(n*p)` or `ceiling(n*p)`) doesn't hold. The mean and most probably value are usually very close to each other (they are each other in the integer limit), but when dealing with small values like we are here, this fails, and this is why the plugin was calculating incorrect values.

**How is this fixed?** Rather than start looking from `n*p` (or the most probable value, which is harder to find), it now starts looking from the observed s1_top, goes over the top of the curve, and down the other side. This guarantees that we'll find where the blue line crosses the curve.

With this modification, the p-value for this event is calculated at 0.3


pdeperio:
This file has been copied manually to ```pax_v6.8.0``` environment for mass production:
```
/project/lgrandi/anaconda3/envs/pax_v6.8.0/lib/python3.4/site-packages/pax-6.8.0-py3.4.egg/pax/plugins/interaction_processing/S1AreaFractionTopProbability.py
```
corresponding to hax release v2.2.0 (calculated in ```PositionReconstruction``` treemaker v0.19).

PR #662, opened 2018-01-06
Update  time-dependent 3D FDC maps for SR1

jingqiangye:
**Update**: Use Kr data up to Jan. 2, 2018
**Current**: Kr data up to Oct 2017

With using more Kr data, time-dependent 3d fdc can take care of the charge accumulation effect better, especially for time after October.

Following plot shows how the updated correction differs with current correction. They behave almost the same for period before Oct. But for time after October, since current correction is only based on data by then, it's impossible for it to handle the more severe field distortion due to accumulating charge. As a result, we can find current position moves inwards, which is shown moving outwards in the plot.

Data: background(cs1 < 200 PE)
r_diff_bg


pdeperio:
delete old files?

jingqiangye:
The outdated 3D FDC maps are deleted now.

pdeperio:
Seems only deleted 1 file?

pdeperio:
Manually copied to ```pax_v6.8.0``` environment:
```
/project/lgrandi/anaconda3/envs/pax_v6.8.0/lib/python3.4/site-packages/pax-6.8.0-py3.4.egg/pax/data
```

PR #663, opened 2018-01-17
Don't include inactive PMTs in S1 pattern fit

JelleAalbers:
The S1 pattern fit calculation currently doesn't account for inactive PMTs when comparing the observed pattern to the MC map. This means the pattern fit values are unnecessarily bad, and the power of S1 pattern fit cut is much reduced.

This fixes the bug. I checked using a simulated 1400 PE S1 from an interaction at (x=0, y=0, z=-50):

	Current master:
		s1_pattern_fit = 803
		s1_pattern_fit_hits = 343
		
	After fix:
		s1_pattern_fit = 314
		s1_pattern_fit_hits = 237

At low-energy the effects might be a bit less dramatic, since the S1 hitpatterns are far more stochastic.

As an 'interesting' side effect, this currently changes @sanderbreur's hitpattern-based "saturation" correction (which is unused at the moment) to also fill in the area of dead PMTs using the MC LCE map. Thus the S1 light yield map should be more uniform after applying the correction: would be interesting to check if this is true once we remake the S1 LCE map. Still, we probably don't want this: it means you can't use the saturation correction alongside the current S1 spatial correction. This can be easily fixed, but my mind has stopped working and just fixing this bug is more important.

Thanks to @coderdj for spotting the problem.


feigaodm:
Thanks @JelleAalbers! For SR1 analysis purpose, we shall modify [HAX treemaker](https://github.com/XENON1T/hax/blob/25f9f3f77774862b9af0defc0e7de71c8e70067a/hax/treemakers/posrec.py#L243) to include this list as well. I guess the gains can be extracted from rundb so that we can consider run dependence as well.  @skazama , can you modify the corresponding code in HAX and we can regrow minitrees right afterwards.

coderdj:
@skazama no need. I already modified it. We'll merge that in with the new s2 pattern variable hopefully today.

PR #664, opened 2018-01-17
Update S2 xy map in SR1 config

pdeperio:
Forgot to update S2 xy correction map in ```SR1_parameters.ini``` to v2.2 (as in [hax.ini](https://github.com/XENON1T/hax/blob/v2.3.1/hax/hax.ini#L125) now). This config is mostly used in simulations now, but perhaps we should also use for hax [here](https://github.com/XENON1T/hax/blob/v2.3.1/hax/treemakers/posrec.py#L60) if not defined [there](https://github.com/XENON1T/hax/blob/v2.3.1/hax/hax.ini#L122) already, for example ```s2_patterns_file``` which has changed in ```XENON1T.ini``` since the ```pax_v6.8.0``` reprocessing, and we need to be consistent now for recalculations.

### DIFF
--- a/pax/config/SR1_parameters.ini
+++ b/pax/config/SR1_parameters.ini
@@ -21,7 +21,7 @@ drift_field =                         82 * V /cm
 
 # Light distribution
 s1_light_yield_map =                  's1_xyz_XENON1T_kr83m-sr1_pax-664_fdc-adcorrtpf.json'
-s2_light_yield_map =                  's2_xy_map_v2.1.json'
+s2_light_yield_map =                  's2_xy_map_v2.2.json'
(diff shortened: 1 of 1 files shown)


pdeperio:
This PR fixes ```s2_light_yield_map```.

I was referring to a potential solution to the current development of ```s2_pattern_fit``` in hax, which you had to [hardcode for now](https://github.com/XENON1T/hax/blob/2731fd5a1ea61019280991d5cd7d003fcb06b4bf/hax/treemakers/posrec.py#L84). So was thinking we could load ```SR1_parameters``` there instead of ```XENON1T``` to grab this file.

coderdj:
Ah, yeah it wouldn't be a bad idea to take snapshots of each .ini as used for each result and keep for such purposes. But really we should not be correcting in hax in this way (or we should reorganize such that such corrections fit into the data flow better).

PR #665, opened 2018-01-28
Change S1 Decay Time in Sim

zhut19:
Change S1 decay time to match data
Step compare data with fax
width_compare_1 08

Maximize Log Likelihood
maximum_ll


feigaodm:
Thanks @zhut19, Looks solid! The mean value should be directly useable.
We'd like to evaluate the undertainty of this parameter as well, from you likelihood curve. I think your likelihood value could be wrong. Can you write down you likelihood?
The easist would be to assume gaussian error bars on sim and data, and calculate chi^2/ll values.

zhut19:
@feigaodm from data took data means as Y0, fax meas as Y
log likelihood = sum( [-(Y-Y0)**2]  )
(As I don't know the uncertainties on the means)

feigaodm:
@zhut19 Yeah, I think you derived the quantiles from simulated and real data. You can calculate the uncertainty on the quantiles, I don't remember a formular to do this, at least you can use bootstrap technique. (Or you can google a way to calculate it, I did it but couldn't find a easy solution). Then you can use the uncertainty to recalculate your likelihood.

pdeperio:
how does this affect @JelleAalbers's [S1 width matching](https://xe1t-wiki.lngs.infn.it/lib/exe/detail.php?id=xenon%3Axenon1t%3Aanalysis%3Asciencerun1%3Awaveform_simulator_summary_170629&media=xenon:xenon1t:aalbers:s1_waveform_shape:s1_width_comparison.png) (in [waveform simulator summary note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:sciencerun1:waveform_simulator_summary_170629))?

JelleAalbers:
Thanks for the study @zhut19. I used the simulation code in https://github.com/JelleAalbers/xeshape for my matching, which is slightly different than fax (but much faster, so waveform template matching in several dimensions is more practical). I tried to make them as similar as possible, but I guess some residual difference remains, causing this discrepancy. When I have time I'll try to compare the two in more detail and hopefully reconcile them. For now of course it's agreement between fax and data that counts, so this adjustment is very welcome.

PR #670, opened 2018-02-03
Update 3D FDC map for 4th period

jingqiangye:
Only the 3D FDC map for 4th period(10/07/2017- 01/30/2018) is changed due to the new acquired Kr data, the rest 3 periods are still using previous maps.

Following two plots are showing low energy ER background(cs1 < 200 PE, inside ER band)  radial distribution. We can find after 3D FDC update, the peak is moving closer to expected TPC boundary. We still observe that even after 3D FDC update, the peak position is still less than 47.9 cm. A potential explanation for that is we have less Kr in later stage due to Rubidium decay, thus we're actually  focus less on the later stage, which has more field distortion. Currently, this effect is very small(less than 0.1 cm), considering the position resolution. It should be improved in the future.
plot_before_update
plot_after_update


feigaodm:
looks like it's a little off 47.9, any explanation?

jingqiangye:
A potential explanation for that is we have less Kr in later stage due to Rubidium decay, thus we're actually focus less on the later stage, which has more field distortion.

feigaodm:
@jingqiangye Please delete older files as requested and we can merge.

jingqiangye:
Done. Shall I change Correction minitree version? Because I remember Dan said e-lifetime needs to be update as well, so I'm not sure if I shall change the version number right now.

pdeperio:
We'll take care of hax versioning: https://github.com/XENON1T/hax/pull/212

Issue #675, opened 2018-02-16
No module pax.plugins.interaction_processing.S1AreaFraktionTopProbability


XeBoris:
Dear all,

While running cax in the slow control screen session I see this error message (taken out from cax.log on xe1t-datamanager):

```
cax_v5.2.1 - 2018-02-16 11:33:03,188 [ERROR] No module named 'pax.plugins.interaction_processing.S1AreaFractionTopProbability'
Traceback (most recent call last):
  File "/home/xe1tdatamanager/cax_fl_test/cax/main.py", line 147, in main
    task.go(args.run)
  File "/home/xe1tdatamanager/cax_fl_test/cax/task.py", line 65, in go
    self.each_run()
  File "/home/xe1tdatamanager/cax_fl_test/cax/tasks/corrections.py", line 78, in each_run
    {'$set': {self.key: self.evaluate(),
  File "/home/xe1tdatamanager/cax_fl_test/cax/tasks/corrections.py", line 138, in evaluate
    gains = self.get_gains(timestamp)
  File "/home/xe1tdatamanager/cax_fl_test/cax/tasks/corrections.py", line 156, in get_gains
    hax.init(pax_version_policy='loose', main_data_paths=[])
  File "/home/SHARED/anaconda3/envs/pax_head/lib/python3.4/site-packages/hax-2.4.0-py3.4.egg/hax/__init__.py", line 71, in init
    minitrees.update_treemakers()
  File "/home/SHARED/anaconda3/envs/pax_head/lib/python3.4/site-packages/hax-2.4.0-py3.4.egg/hax/minitrees.py", line 147, in update_treemakers
    __import__('hax.treemakers.%s' % module_name, globals=globals())
  File "/home/SHARED/anaconda3/envs/pax_head/lib/ROOT.py", line 301, in _importhook
    return _orig_ihook( name, *args, **kwds )
  File "/home/SHARED/anaconda3/envs/pax_head/lib/python3.4/site-packages/hax-2.4.0-py3.4.egg/hax/treemakers/posrec.py", line 10, in <module>
    from pax.plugins.interaction_processing.S1AreaFractionTopProbability import s1_area_fraction_top_probability
  File "/home/SHARED/anaconda3/envs/pax_head/lib/ROOT.py", line 301, in _importhook
    return _orig_ihook( name, *args, **kwds )
ImportError: No module named 'pax.plugins.interaction_processing.S1AreaFractionTopProbability'
```

[...]

Issue #674, opened 2018-02-15
Latest numpy required (really?) but not yet available from conda


JelleAalbers:
In https://github.com/XENON1T/pax/pull/673 we started using a numpy >= 1.12 feature, but numpy 1.11 is the latest available version with python 3.4 through the regular conda channels. I see you solved this by adding 'pip install numpy' to the travis guidelines. However, this doesn't do the detailed dependency checks that conda does (for example, numba relies on numpy and might now break something else in pax).

Hopefully anaconda's numpy will update soon. In the meantime, could https://github.com/XENON1T/pax/pull/673 be implemented without using the signature argument? I'd rather not replace conda numpy with pip numpy on the eventbuilder, so I can't use the  latest pax at the moment.

feigaodm:
Thanks for the report,  can @zhut19 help to fix it?

zhut19:
Thanks @JelleAalbers, I can remove the signature kwarg by commenting out the np vectorize section of the pull request, but I wasn't able to find a way using np vectorization here without this kwarg.

And this pull request would still decrease the interpolation time without this np vectorization. So do @JelleAalbers @feigaodm we agree on go back a step for now?

feigaodm:
@zhut19 Can you keep the cKDTree in the code so we still expect some speed improvement?

zhut19:
Yes, the cKDTree can still be in there.

feigaodm:
ok, then I agree, please make a PR for it.

feigaodm:
fixed by #676

PR #676, opened 2018-02-16
Retract in interpolation

zhut19:
In response to issue #674 and problem raised in issue #675, we would comment out the np vectorization part from pr #672. And keeps the cKTree part


JelleAalbers:
Thanks Tianyu! Let's re-enable this as soon as the new numpy becomes available in Anaconda.

PR #681, opened 2018-04-04
Attempt TravisCI caching and pinning versions to Midway pax_head

tunnell:
To ensure that we are able to always know what versions of packages are with each pax release, I'm testing pinning versions of dependencies.  This will get tested by Travis.  If this works, we can rely on services like pyup (see #680)


setup.py
```diff
@@ -16,16 +16,8 @@
...
-# For some reason h5py is often not seen by pip if it was installed by conda...
-# so check for h5py presence manually, and remove it from requirements if already found.
-try:
-    import h5py
-except ImportError:
-    pass
-else:
-    del requirements[requirements.index('h5py')]
-
+
```
tunnell: __[Flake8]__

[W293] blank line contains whitespace

tunnell:
Note that this also allows for TravisCI cachine

tunnell:
See #686.  Instead of working my way up, working my way down.

PR #683, opened 2018-04-04
Replace fromstring to frombuffer

tunnell:
frombuffer is now preferred within numpy when dealing with binary data.    This solves "DeprecationWarning: The binary mode of fromstring is deprecated, as it behaves surprisingly on unicode inputs. Use frombuffer instead"


tunnell:
@feigaodm @JelleAalbers thoughts?  needs approval, ready to merge

JelleAalbers:
I'm surprised numpy deprecated this. `frombuffer` has the problem that if you pass it an immutable object (such as a `bytes`) the resulting array will also be readonly. That can cause some interesting failures; the solution is probably to copy things into a bytearray first. But since the test pass, we could just wait if something will go wrong now...

feigaodm:
haha, @JelleAalbers sorry if I merged too fast because I see @tunnell was doing a lot of commits and it might be urgent. But as you said, we will see if it brings new problems and fix them

tunnell:
Also, it's mainly the XED thing that we should probably deprecate anyways at some point, even though it's used for all tests

PR #701, opened 2018-04-05
Split dependencies into upgradable and not upgradable

tunnell:
About half of our dependencies cannot have their versions changed without breaking pax.  After testing in my own fork, I've been able to identify which  can be upgrade and which cannot.  The ones that cannot will be marked as ignore with pyup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ addons:
       - libboost-all-dev
       - build-essential
       - libc6-dev
+      - snappy
 
 matrix:
     include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,20 @@ sudo: false
 
 matrix:
     include:
+        - python: 2.7
+          env: PYTHON=2.7 ROOT=5.34.32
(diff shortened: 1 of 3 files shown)


tunnell:
After this, I'd setup pyup.

feigaodm:
neat, so we disable support to python2.7 totally.

tunnell:
Oh right, let me readd.  Uh oh... god I hope that doesn't break it again....

feigaodm:
I think python2.7 is more painful...

tunnell:
Good catch, let's see if it builds.

tunnell:
@JelleAalbers what do you think about dropping 2.7 support?  Given plans anyways. If we do that, than I can move on to seeing what parts of pax we can ax.

JelleAalbers:
@tunnell dropping py2 is fine, nobody is using it and many other projects are dropping py2 very soon too. It's always been a 'best effort' thing, and if you really need py2 you now have many pax releases you can go back to.

PR #702, opened 2018-04-16
Strip requirements versions in setup.py

JelleAalbers:
This changes setup.py to allow `pip install pax` again without necessarily installing every single package version we set in requirements.txt. Without this, #701 basically prevents pax from co-existing with any other package that uses similar or slightly milder restrictions. Pinning versions in  install_requires is also officially frowned upon by the python overlords: https://packaging.python.org/discussions/install-requires-vs-requirements/ 

For automated deployments and repeatable builds we can still use `pip install -r requirements.txt`, which will install only the specific pinned versions that are tested by Travis to work. 

Populating install_requires from requirements.txt at all is actually considered somewhat bad practice, but the alternative (maintaining two separate lists of dependencies manually) is perhaps even more unsavory. Maybe the new generation of packaging tools (pipenv) solves this problem, but at least I'm not ready to switch to that now.

(@tunnell and I had a similar discussion on this in another project, this just copies the solution he eventually proposed.)


JelleAalbers:
I'll merge this to try and trigger a new pax deployment; there was a failure during installation of matplotlib of the last one.

PR #705, opened 2018-05-11
lowering tight coincidence to 2-fold

feigaodm:
This PR lowers the tight coincidence coincidence to two-fold in order to improve the acceptance of low-E events, as requested by @mcfatelin 

I also included a S1width (range_90p_area) in the classification, which should replace the [S1Width cut in LAX](https://github.com/XENON1T/lax/blob/master/lax/lichens/sciencerun0.py#L577).

### DIFF
--- a/pax/config/XENON1T.ini
+++ b/pax/config/XENON1T.ini
@@ -313,9 +313,6 @@ s2_afterpulse_types = {
 max_gap_size_in_cluster = 3.50 * us
 max_gap_size_in_s1like_cluster = 350 * ns
 
-# s1 width/rist_time bound, in ns
-s1_width_threshold = 300
-s1_risetime_threshold = 70
 
 [NaturalBreaksClustering]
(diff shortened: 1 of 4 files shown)


jhowl01:
At RPI @feigaodm started to explore this, and I did some checks on acceptance of S1s. We decided at the time to hold off, since the pax version for SR1 was frozen at 6.8.0. This exploration motivated the splitting of the tight_coincidence_window into "right" and "left" in #644 .

The plot below compares the nominal ~6.8.0 acceptance (at that time) with acceptances from processing with a tight-coincidence requirement of 2 hits (instead of 3). For the lower hit requirement, we also tried reducing the left-of-center window, to try to improve noise-coincidence rejection. If we switched from 3 to 2, and kept the tight_coincidence_window_left at 50 ns, we would see an effect like going from the blue to the purple points.

varying_tight_coincidence_requirement

PR #712, opened 2018-07-13
clustering and desaturation upgrades

zhut19:
Here we propose several upgrades tailored for high energy peaks.
Detailed information is documented in [clustering upgrades](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:sim:notes:tzhu:high_energy_peak_clustering_amendment:multipeakseparation) and [desaturation upgrades](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:chiara:pax_desaturation).


tunnell:
Looks nice, especially the S2 tails work

pax/plugins/peak_processing/BasicProperties.py
```diff
@@ -86,10 +87,10 @@ def transform_event(self, event):
             # Don't weigh negative samples for computation of center of gravity
             weights = np.clip(w, 0, float('inf'))
             if not np.sum(weights) > 0:
-                self.log.warning("Sum waveform of peak %d-%d (%0.2f pe area) sums to a nonpositive value... unusual!"
-                                 " Cannot align peak's sum waveform, storing zeros instead." % (peak.left,
+                self.log.debug("Sum waveform of peak %d %d-%d (%0.2f pe area) sums to a nonpositive value... unusual!"
+                                 " Cannot align peak's sum waveform, and will be deleted." % (peak_i, peak.left,
```
tunnell: __[Flake8]__

[E127] continuation line over-indented for visual indent

pax/config/XENON1T.ini
```diff
@@ -333,6 +334,13 @@ min_height = 25 #pe
 # Minimum min(pheight eak left of loc min, height of peak right of loc min)/height of loc min
 min_ratio = 4
 
+# [LocalMinimumClustering.TailSeparation]
```
feigaodm: delete the lines that doesn't add information

pax/plugins/peak_processing/LocalMinimumClustering.py
```diff
@@ -116,34 +116,77 @@ def split_peak(self, peak, split_points):
...
+
+    # We look for the last minima between big peaks
+    good_minima = [int(partial_xindex[minima[ix]]+left) for ix in range(len(minima))
+                   if ((_w[maxima][ix+1] - _w[minima][ix] > min_height)
+                       & any(_w[maxima][:ix+1] - _w[minima][ix] > min_height))]
+
+    return good_minima
+
+
+class TailSeparation(LocalMinimumClustering):
```
feigaodm: Add some description on this function

pax/plugins/signal_processing/DesaturatePulses.py
```diff
@@ -27,71 +27,117 @@ def transform_event(self, event):
...
+                conv = np.ones(self.config['convolution_length'])/self.config['convolution_length']
+                v = np.convolve(conv, v, mode='same')
+                dv = np.diff(v, n=1)
+                # Choose +2 pe/ns instead 0 to avoid ending on the flat waveform
+                turning_point = np.where((np.hstack((dv, -10)) > 2) & (np.hstack((10, dv)) <= 2))[0]
+
+                if len(turning_point) > 0:
+                    end = last_saturated + turning_point[0]
+
+                # Determine where to stop correction
```
feigaodm: clean the lines if we don't need to leave them their for reference

zhut19:
Thanks @sanderbreur 
 * We are thinking about checking 'tail_cutoff_ratio' using simulated waveform once we can match the waveforms. Not sure how to check it with processing data, except looking at the energy resolution.

 * We are checking the 'large_after_pulsing_channels' list changes between sr0, sr1 and after sr1 with pmt database. We can come up a strategy after checking.

 * We ran into some bug with a peak that set to zeros, and to by pass it, we added those line to actually delete the peaks. And we think those ~0.01 pe peaks won't matter anyway.

 * Line 123 is deleted

 * Added a condition to exclude S1 from TailSeparation.

 * Will add int16 to float 64 to the DesaturatePulses class description

pax/plugins/peak_processing/LocalMinimumClustering.py
zhut19: More description are added to the function, but a relative short one for the class here.

chiaracapelli:
@sanderbreur The list of 'large_after_pulsing_channels' is checked and updates with the PMT database. We are using the 4 datasets during SR1 in the config/pmt_afterpulses, so in principle we don't have to update the list if we use SR1 data. We can update it including SR0 if we want to use also that.

sanderbreur:
Chiara, cool, would be nice to have the list later also for sr0, else sr0
reprocessing will give problems.

chiaracapelli:
unknown

We reprocessed one dataset and checked the s1s and s2s relative difference between the old processed and this new version in 1T FV. The data are mostly in the region 300-1500 keV.

PR #714, opened 2018-07-19
Push Git tag before actual commits

pdeperio:
Noticed some weird behavior with deployHQ where new release environment was not being created on CVMFS. Hypothesis: Do ```git push --tag``` before actual commits so DeployHQ knows about the new tag in order to build new Anaconda environment. Not sure if it'll work (can test at next release), but it shouldn't break the workflow, right @tunnell?


feigaodm:
@tunnell how about this PR?

PR #715, opened 2018-07-19
Move DB examples to processing repo (formerly "Midway environment creation script")

pdeperio:
Midway conda environment and xenon1t package automatic installation script called by DeployHQ. Change from local copy is the cloning of ```pax_head``` for new pax release environments instead of trying to create them from scratch, which was causing a lot of dependency problems as packages evolve over time. (This is still not the most sustainable since ```pax_head``` may change too, though we should be keeping track [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:cmp:computing:conda_log) now.)


pdeperio:
@tunnell ok moved to https://github.com/XENON1T/processing/pull/61. please merge that and this (now deletes the DB notebook examples).

PR #717, opened 2018-08-06
adding coincidence levels in current pax versions

feigaodm:
This is to add a few variables for PMT coincidence analysis, copying @pdeperio 's fix in [pax_v6.8.0.1](https://github.com/XENON1T/pax/releases/tag/v6.8.0.1).


feigaodm:
Fixes https://github.com/XENON1T/pax/issues/718

tunnell:
Uhh, there's an issue.  How does this even work?  One sec.

tunnell:
`tight_coincidence_thresholds = np.zeros(5, dtype=np.int16)  # various tight coincidence levels`

That's not how you define things in datastructure.  Look at the other np arrays.  You have to initialize the values in the plugins.  You also want just the basic np.array type such that the ROOT code knows how to produce the ROOT C++ class.

feigaodm:
I was trying to copy what was done for peak width variables, like
 'range_area_decile = np.zeros(11, dtype=np.float)'

pax/datastructure.py
```diff
@@ -278,6 +278,7 @@ def contributing_channels(self):
     #: Number of channels that have a hit maximum within a short (configurable) window around the peak's sum
     #: waveform maximum.
     tight_coincidence = INT_NAN
+    tight_coincidence_thresholds = np.zeros(5, dtype=np.int16)  # various tight coincidence levels
```
tunnell: This should be `np.array([], dtype=np.int16)` otherwise the ROOT class doesn't know how to write the C++ schema.

tunnell: Look at other array fields and how they are set within plugins.

tunnell:
Oh really?  Odd.  I thought we had very strict requirements on what datatypes could be converted.  OK, it's too late right now, maybe bad to review.

pax/datastructure.py
pdeperio: https://github.com/XENON1T/pax/blob/6bc98d8fb41b600308aed44c423c8a682c6f7ee9/pax/datastructure.py#L300 ?

tunnell:
Due to #718, I would still try what I suggest.  Look at area_per_channel for example.  I'm not sure why the width code works, but this stuff is very finicky so it's worth trying to add the array like done for the others because that's clearly the source of the issue.

PR #725, opened 2019-03-01
Segregate hit smoothing

jhowl01:
#712 included the addition of additional [smoothing and splitting](https://github.com/XENON1T/pax/pull/712/files#diff-5d2794434093c23639c8c8f1e5044cb2R203) when `find_intervals_above_threshold()` is called in the hitfinder. This added some numpy functions, which necessitated the removal of the numba speed boosting. This function is also caused during the ZLE phase of [fax](https://github.com/XENON1T/pax/blob/master/pax/plugins/ZLE.py#L55), which has since slowed down significantly. Unfortunately, since the numba boosting couldn't work on the current function, I didn't see a fix besides duplicating a good amount of code 🤷‍♂️ . In this PR, the old version of the function is instead called in `ZLE.py`, restoring fax's speed.

### DIFF
--- a/pax/dsputils.py
+++ b/pax/dsputils.py
@@ -229,3 +229,50 @@ def find_intervals_above_threshold(w, threshold, result_buffer):
 
     n_intervals = current_interval      # No +1, as current_interval was incremented also when the last interval closed
     return n_intervals
+
+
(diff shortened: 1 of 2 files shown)


feigaodm:
I didn't see flake8 message:) will merge shortly

Issue #703, opened 2018-04-17
Deprecated matplotlib function in PatternFitter


pelssers:
I recently had to install Pax in a Python 3.5 environment. The problem I encountered was that in the PatternFitter we use the matplotlib module _cntr which is deprecated since matplotlib 2.2.0.

We used this module instead of matplotlibs contour() function because it traces contours without making an actual plot. To work with the latest matplotlib we need to rewrite this part of PatternFitter using matplotlibs contour() function.

feigaodm:
@pelssers Would you like to fix it with a PR, maybe that helps to understand what's the issue :)

JelleAalbers:
It seems we're using _cntr to find an array of (x,y) pairs describing the contour (https://github.com/XENON1T/pax/blob/master/pax/PatternFitter.py#L268). We could use http://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.find_contours from scikit-image to do it instead; it might even simplify the code. I'm fine with adding scikit-image as a dependency.

PR #734, opened 2019-04-24
Allow import failure for matplotlib._cntr

JelleAalbers:
This is a workaround for #703. Currently pax does not run with matplotlib > 2.2.0 due to using the private matplotlib function `_cntr` that has been removed.  This will allow pax to import and run even with a newer matplotlib -- although a warning message will be printed, since confidence tuples from TopPatternFit position reconstruction cannot be produced.

The main use of this is making pax import-able in more up-to-date python environments, such as those used for pax->strax format conversion. Alternatively you could monkeypatch matplotlib to achieve this:
```
import matplotlib
matplotlib._cntr = None
import pax
```


pax/PatternFitter.py
```diff
@@ -250,7 +254,8 @@ def minimize_gof_grid(self, center_coordinates, grid_size, areas_observed,
         # Store (dx, dy) for each CL for output
         confidence_tuples = []
 
-        if cls is not None and n_dim == 2:
+        if cls is not None and n_dim == 2 and _cntr is not None:
+
```
tunnell: __[Flake8]__ [W293] blank line contains whitespace ([link](https://sider.review/gh/repos/21694081/pulls/734#issue-4004362))

<sub>posted by Sider</sub>

JelleAalbers:
@feigaodm Any objections to this? It would enable pax to be run in more up-to-date python environments.

PR #737, opened 2020-03-29
Update tight-coincidence definition

jhowl01:
This changed the definition of peaks' tight-coincidence. Previously, tight-coincidence was the number of hits whose **maximum amplitude occurs within `tight_coincidence_window` of the peak's maximum amplitude**. In this PR, it is changed to the number of hits whose **mean occurs within `tight_coincidence_window` of the peak's mean**. For peaks of a given size, this should reduce the variance of the tight_coincidence statistic. The plot shows the improvement in S1 acceptance for a few window sizes (window_left = window_right = window). We also expect reduced pileup of random hits for a given window, since the window isn't smeared out as much.

The PR also saves the tight-coincidence number for a few symmetric window sizes (30ns, 40ns, 50ns), while using the left and right windows from the config for classification.

modify_tight_coincidence_200329

### DIFF
--- a/pax/plugins/peak_processing/BasicProperties.py
+++ b/pax/plugins/peak_processing/BasicProperties.py
@@ -1,7 +1,7 @@
 import numpy as np
 # import numba
 
-from pax import plugin, dsputils
+from pax import plugin, dsputils, units
 
 
 class BasicProperties(plugin.TransformPlugin):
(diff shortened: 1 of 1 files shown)


jhowl01:
To decide the impact of this, maybe @zhut19 can check the lone-S1 rate by processing a bit of data on this branch.

zhut19:
We see the lone-s1 doubled given this pr shown in the following table, we believe the extra lone-s1s are very likely to be pileup as real s1 rate is not expected to double like this.

| Lone S1 | w/o this pr  |  w/ this pr |
|---|---|---|
|Both Arraies|  9.6 Hz |  15.7 Hz  |
|Only Bottom|  2.8 Hz |  4.6 Hz   |

feigaodm:
thanks for the effort! Since it's not helping, let's close it.

PR #741, opened 2020-08-26
SR1 S1 LCE for pax version above 6.10.0

zhut19:
# Please read before creating a Pull Request

Before you start a pull request (PR), please note that pax is public. Only the technical details of the code can be mentioned directly on GitHub.  If the analysis details are necessary for the PR, please make a wiki note and link it.  The pull request will focus on the technical aspects rather than the motivation for the feature request.


feigaodm:
I guess a PR in hax is needed as well if we merge this PR, right? Otherwise it does nothing, no?

